Issue #85, opened 2019-10-08
Docs should contain API endpoints for interacting with dapr and how dapr interacts with user code


AaronCrawfis:
For example dapr service invocation is triggered by:
`http://localhost:<DAPR_HTTP_PORT>/v1.0/invoke/<id>/method/<method-name>`

and reaches out to the app by:

`http://localhost:<app-port>/<method-name>`

jocko-wowza:
Is there somewhere we can get details like @AaronCrawfis is asking about.   I have a configuration that is using Redis as the event store, but I can't find any documentation on how to trigger an event in this configuration.

jocko-wowza:
I believe I found the answer in the sample code repository.   It's not in line with the documentation, so it was difficult to find, and you have to read the code to determine the actual endpoints.

Reading through the sample code here: 
https://github.com/dapr/samples/tree/master/5.bindings

The Node app code actuates the trigger by calling this DAPR endpoint:
`http://localhost:{}/v1.0/bindings/sample-topic".format(dapr_port)`

Which appears to use this conventional structure:
`http://localhost:[DAPR-port-number]/v1.0/bindings/[event-name]`

amanbha:
Addressed in PR https://github.com/dapr/docs/pull/381

PR #237, opened 2019-11-11
Added endpoint details for event triggers.

jocko-wowza:
# Description

Added a section to the inbound bindings documentation that describes how to trigger an event.

While referencing the example briefly in the head of the inbound bindings documentation, the doc was not explicit about how to actually trigger an event. 

## Issue reference

Issue #85 

## Checklist

* [N/A] Code compiles correctly
* [N/A] Created/updated tests

### DIFF
--- a/howto/send-events-with-output-bindings/README.md
+++ b/howto/send-events-with-output-bindings/README.md
@@ -37,7 +37,7 @@ All that's left now is to invoke the bindings endpoint on a running Dapr instanc
 We can do so using HTTP:
 
 ```
-curl -X POST -H  http://localhost:3500/v1.0/bindings/myEvent -d '{ data: { "message": "Hi!" } }'
+curl -X POST -H  http://localhost:3500/v1.0/bindings/myEvent -d '{ "data": { "message": "Hi!" } }'
 ```
 
 As seen above, we invoked the `/binding` endpoint with the name of the binding to invoke, in our case its `myEvent`.<br>
--- a/howto/trigger-app-with-input-binding/README.md
+++ b/howto/trigger-app-with-input-binding/README.md
@@ -81,3 +81,13 @@ res.status(500).send()
 
 ### Event delivery Guarantees
 Event delivery guarantees are controlled by the binding implementation. Depending on the binding implementation, the event delivery can be exactly once or at least once.
+
+## 3. Triggering the Event 
+
+Now you can trigger the event from within your code using the DAPR bindings endpoint, which is structured like this:
+
+```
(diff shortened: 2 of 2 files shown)


yaron2:
He @jocko-wowza, first of all, thanks for this PR!

However, have you seen this how-to for *output* bindings?
https://github.com/dapr/docs/tree/master/howto/send-events-with-output-bindings

This how-to describes output bindings, while the one you extended is solely focused on input bindings.

jocko-wowza:
Maybe I mis-typed something, but what I was reacting to was Input Bindings, not Output bindings, and that it was hard to find the right endpoint to send a binding trigger to when operating in the context of an Input Binding.

jocko-wowza:
I just realized that pull request also includes the other change we are discussing.  Apologies for that.

yaron2:
> I just realized that pull request also includes the other change we are discussing. Apologies for that.

Lets merge the other PR first and then continue to discuss this one..

jocko-wowza:
So with your comments above and the documentation, am I to understand that the articles:

- https://github.com/dapr/docs/tree/master/howto/trigger-app-with-input-binding
- https://github.com/dapr/docs/tree/master/howto/send-events-with-output-bindings

Are two halves of a workflow where you define an input-binding to react to an event trigger, and then call an output-binding to trigger that event?

yaron2:
> So with your comments above and the documentation, am I to understand that the articles:
> 
> * https://github.com/dapr/docs/tree/master/howto/trigger-app-with-input-binding
> * https://github.com/dapr/docs/tree/master/howto/send-events-with-output-bindings
> 
> Are two halves of a workflow where you define an input-binding to react to an event trigger, and then call an output-binding to trigger that event?

Not necessarily that specific event, but yes, you are right: the first one teaches you on how to listen for input binding events, and the second one complements it by teaching you how to invoke output bindings.

The bindings *sample* ties the two together.

jocko-wowza:
Okay ... So as feedback, I read them as being two different conceptual forms of bindings.   Perhaps the input binding documentation could reference the Output binding documentation at the end instead.   If you want I can write it up and do another pull request.

yaron2:
> Okay ... So as feedback, I read them as being two different conceptual forms of bindings. Perhaps the input binding documentation could reference the Output binding documentation at the end instead. If you want I can write it up and do another pull request.

Sounds good. it'd be best to insert a small section at the end of the input bindings how-to that links up to the output bindings one as continuity.

jocko-wowza:
Fixed in another PR. Closing

PR #264, opened 2019-11-28
Updated error on password placement in setup-redis.md

lewisbenge:
# Description

Corrected typo in documentation stating password should be placed in redisHost when it should be placed in redisPassword


yaron2:
Thanks!

PR #273, opened 2019-12-07
Describe the sequence of events when running dapr run

LMWF:
# Description

Describe the sequence of events when running `dapr run`.

Let me know if there's a better location for this doc.

## Issue reference

Please reference the issue this PR will close: #_254_

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [x] Extended the documentation

### DIFF
--- a/walkthroughs/darprun.md
+++ b/walkthroughs/darprun.md
@@ -3,8 +3,8 @@
 The doc describes the sequence of events that occur when `dapr run` is executed in self hosting mode (formerly known as standalone mode).  It uses [sample 1](https://github.com/dapr/samples/tree/master/1.hello-world) as an example.
 
 Terminology used below:
-- Dapr CLI – the Dapr command line tool.  The binary name is dapr (dapr.exe on Windows)
-- Dapr runtime – this runs alongside each app.  The binary name is daprd (daprd.exe on Windows)
+- Dapr CLI - the Dapr command line tool.  The binary name is dapr (dapr.exe on Windows)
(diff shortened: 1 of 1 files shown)


walkthroughs/darprun.md
```diff
@@ -0,0 +1,51 @@
...
+```
+dapr run --app-id nodeapp --app-port 3000 --port 3500 node app.js
+```
+
+First, the Dapr CLI creates the `\components` directory if it does not not already exist, and writes two component files representing the default state store and the default message bus: `redis.yaml` and `redis_messagebus.yaml`, respectively.  [Code](https://github.com/dapr/cli/blob/d585612185a4a525c05fb62b86e288ccad510006/pkg/standalone/run.go#L254-L288).
+
+*Note as of this writing (Dec 2019) the names have been changed to `statestore.yaml` and `messagebus.yaml` in the master branch, but this change is not in the latest release, 0.3.0*.  
+
+
+Components describe configuration (e.g. how the Dapr runtime should communicate to the pub/sub).  The pub/sub itself must be created prior.  For example, Redis is launched as a container when running `dapr init`.  If these component files already exist, they are not overwritten.  This means you could overwrite `statestore.yaml`, which by default uses Redis, with a content for a different statestore (e.g. Mongo) and the latter would be what gets used.  If you did this and ran `dapr run` again, the Dapr runtime would use the specified Mongo state store.
```
amanbha: Can it be changed to something like:
yaml files in components directory contains configuration for various Dapr components (e.g. statestore, pubsub, bindings etc.).  The components must be created prior to using them with Dapr, for example, redis is launched as a container when running `dapr init`.

PR #275, opened 2019-12-08
Corrected local development instructions

trondhindenes:
# Description
Now that the filenames in `components` are generic (https://github.com/dapr/cli/pull/218), this PR attempts to update documentation accordingly. Also attempts to specify that the filenames inside `components` should not be manipulated, since `dapr/cli` has strict expectations to what those names should be.

## Issue reference
N/A

## Checklist
* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Extended the documentation


LMWF:
Thanks for your contribution!

PR #286, opened 2019-12-30
Use checkmarks to show bindings support

dev-drprasad:
# Description

It took me a while to understand what `V` stands for in supported bindings table. I thought using checkmarks makes more sense

Feel free to ignore PR for any other reason


yaron2:
Thanks!

PR #341, opened 2020-02-10
Update security.md

yaron2:
Updates with mTLS details.

### DIFF
--- a/concepts/security/security.md
+++ b/concepts/security/security.md
@@ -29,6 +29,14 @@ This can be done for both Kubernetes and self hosted modes.
 
 Specific details for how to do that can be found [here](../../howto/configure-mtls).
 
+### mTLS in Kubernetes
+
+<a href="https://ibb.co/LpyLv7g"><img src="https://i.ibb.co/bdVS79v/Screen-Shot-2020-02-10-at-7-51-51-PM.png" alt="Screen-Shot-2020-02-10-at-7-51-51-PM" border="0"></a>
+
+### mTLS Self Hosted
+
+<a href="https://ibb.co/Btny30B"><img src="https://i.ibb.co/t4LqKVD/Screen-Shot-2020-02-10-at-7-56-19-PM.png" alt="Screen-Shot-2020-02-10-at-7-56-19-PM" border="0"></a>
+
 ## Network security
 
 You can adopt common network security technologies such as network security groups (NSGs), demilitarized zones (DMZs) and firewalls to provide layers of protections over your networked resources.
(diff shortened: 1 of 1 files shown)


concepts/security/security.md
```diff
@@ -6,13 +6,24 @@ Dapr sidecar runs close to the application through **localhost**. Dapr assumes i
...
+When root certs are replaced (Secret in Kubernetes mode and filesystem for self hosted mode), Sentry will pick them up and re-build the trust bundle without needing to restart, again with zero downtime to Sentry.
+
+When a new Dapr sidecar inits, it will first check if mTLS is enabled. if it is, an ECDSA private key and certificate signing request are generated and sent to Sentry via a gRPC interface. The communication between the Dapr sidecar and Sentry is authenticated using the trust chain cert, which is injected to each Dapr sidecar by the Dapr sidecar injector.
+
+The secret that hold the root certificates are scoped to the namespace in which the Dapr components are deployed to and are only accessible by the Dapr system pods.
+
+Dapr also support strong identities when deployed on Kubernetes, relying on a pod's Service Account token which is sent as part of the CSR request to Sentry.
+
+Mutual TLS can be turned off by editing the default configuration that is deployed with Dapr via the `spec.mtls.enabled` field.
+
```
youngbupark: maybe it is good to provide exact yaml

yaron2:
Not here, there's another PR coming after that.

youngbupark:
> Not here, there's another PR coming after that.

ok then. it would be good to have a link after that

concepts/security/security.md
yaron2: Now it's in here..

yaron2:
@msfussell since this is a concept doc, I removed the instructions and instead linked to a placeholder which will contain the next PR for the actual instructions in the How-To section.

This is in order to clear ambiguity and prevent duplicate content, leaving the concepts for explaining the concepts and free of implementation details.

msfussell:
It would be good to have a diagram in the concepts to iwth words on how the sentry service works, in the same way that the placement service was done for Actors here https://github.com/dapr/docs/blob/master/concepts/actor/actor_overview.md. This prevents us from having to repeat whet the sentry service does and how it works.

yaron2:
> It would be good to have a diagram in the concepts to iwth words on how the sentry service works, in the same way that the placement service was done for Actors here https://github.com/dapr/docs/blob/master/concepts/actor/actor_overview.md. This prevents us from having to repeat whet the sentry service does and how it works.

When you say "with words", you mean the existing description, yes? so only the diagram.

msfussell:
If asked "how does the dapr sentry service work" sending a link to this concept file would satisfy that ask including saying words along the line of "each dapr instance get a cert generated from the sentry service" etc. It is a significant piece of functionality and there will be many ops people who will want to understand how this works to satisfy security requirements in their organization.

yaron2:
> If asked "how does the dapr sentry service work" sending a link to this concept file would satisfy that ask including saying words along the line of "each dapr instance get a cert generated from the sentry service" etc. It is a significant piece of functionality and there will be many ops people who will want to understand how this works to satisfy security requirements in their organization.

The diagram is needed as well. The words that say that each Dapr instance is getting it's own cert are already there.

Issue #316, opened 2020-02-03
Document serialization for transient and state objects in SDKs


artursouza:
Document behavior of serializers in Java and .Net SDKs.

If there is difference in behavior for current version, document that too.

PR #344, opened 2020-02-11
Document expected serialization in SDKs.

artursouza:
# Description

Document expected serialization in SDKs.
Link to example for Java actors.

## Issue reference

Please reference the issue this PR will close: https://github.com/dapr/docs/issues/316

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/howto/README.md
+++ b/howto/README.md
@@ -45,3 +45,7 @@ Here you'll find a list of How To guides that walk you through accomplishing spe
 ## Configuring Visual Studio Code
 
 * [Debugging with daprd](./vscode-debugging-daprd)
+
+## SDKs
+
+* [Serialize objects](./serialize)
--- a/howto/serialize/README.md
+++ b/howto/serialize/README.md
@@ -1,0 +1,146 @@
+# Serialization in Dapr's SDKs
+
+An SDK for Dapr should provide serialization for two use cases. First, for API objects sent through request and response payloads. Second, for objects to be persisted. For both these use cases, a default serialization is provided. In the Java SDK, it is the [DefaultObjectSerializer](https://dapr.github.io/java-sdk/io/dapr/serializer/DefaultObjectSerializer.html) class, providing JSON serialization.
+
+## Service invocation
+
+```java
+    DaprClient client = (new DaprClientBuilder()).build();
+    client.invokeService(Verb.POST, "myappid", "saySomething", "My Message", null).block();
+```
+
+In the example above, the app will receive a `POST` request for the `saySomething` method with the request payload as `"My Message"` - quoted since the serializer will serialize the input String to JSON.
+
+```text
+POST /saySomething HTTP/1.1
+Host: localhost
+Content-Type: text/plain
+Content-Length: 12
+
+"My Message"
+```
+
+## State management
+
+```java
+    DaprClient client = (new DaprClientBuilder()).build();
+    client.saveState("MyStateStore", "MyKey", "My Message").block();
+```
+In this example, `My Message` will be saved. It is not quoted because Dapr's API will internally parse the JSON request object before saving it.
+
+```JSON
+[
+    {
+        "key": "MyKey",
+        "value": "My Message"
+    }
+]
+```
+
+## PubSub
+
+```java
+  DaprClient client = (new DaprClientBuilder()).build();
+  client.publishEvent("TopicName", "My Message").block();
+```
+
+The event is published and the content is serialized to `byte[]` and sent to Dapr sidecar. The subscriber will receive it as a [CloudEvent](https://github.com/cloudevents/spec). Cloud event defines `data` as String. Dapr SDK also provides a built-in deserializer for `CloudEvent` object.
+
+```java
+  @PostMapping(path = "/TopicName")
+  public void handleMessage(@RequestBody(required = false) byte[] body) {
+      // Dapr's event is compliant to CloudEvent.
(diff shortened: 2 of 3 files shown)


howto/README.md
```diff
@@ -41,3 +41,7 @@ Here you'll find a list of How To guides that walk you through accomplishing spe
 ## Configuring Visual Studio Code
 
 * [Debugging with daprd](./vscode-debugging-daprd)
+
+## SDKs
+
+* [Serialize state and transient objects](./serialize)
```
LMWF: I don't think we should use the term "transient object", it sounds odd.  Maybe "method invocation data/objects"?

howto/serialize/README.md
```diff
@@ -0,0 +1,27 @@
+# Serialization in Dapr's SDKs
+
+An SDK for Dapr should provide serialization for two use cases. First, for transient objects sent through request and response payloads. Second, for objects to be persisted. For both these use cases, a default serialization must be provided. In the Java SDK, it is the [DefaultObjectSerializer](https://dapr.github.io/java-sdk/io/dapr/serializer/DefaultObjectSerializer.html) class.
```
LMWF: Lining up these 2 cases with the set of user-exposed apis that use them will make this more concrete.

howto/serialize/README.md
```diff
@@ -0,0 +1,27 @@
...
+redis-cli MGET "ActorStateIT_StatefulActorService||StatefulActorTest||1581130928192||message
+"This is a message to be saved and retrieved."
+```
+```bash
+ redis-cli MGET "ActorStateIT_StatefulActorService||StatefulActorTest||1581130928192||mydata
+{"value":"My data value."}
+```
+3. Custom serializers must serialize object to `byte[]`.
+4. Custom serializers must deserilize `byte[]` to object.
+5. When user provides a custom serializer, it should be transfered or persisted as `byte[]`. When persisting, also encode as Base64 string. This is done natively by most JSON libraries.
```
LMWF: transferred

howto/serialize/README.md
```diff
@@ -0,0 +1,27 @@
+# Serialization in Dapr's SDKs
```
LMWF: This today should say java sdk.  Custom serialization is not available in the others.

artursouza: This doc is generic. I will mention that Java is the only one implementing this.

howto/README.md
artursouza: Changed it.

howto/serialize/README.md
artursouza: Done.

howto/serialize/README.md
artursouza: done.

howto/README.md
```diff
@@ -45,3 +45,7 @@ Here you'll find a list of How To guides that walk you through accomplishing spe
 ## Configuring Visual Studio Code
 
 * [Debugging with daprd](./vscode-debugging-daprd)
+
+## SDKs
+
+* [Serialize state and request/response objects](./serialize)
```
LMWF: Maybe the better way is to have a terminology section at the top, saying "in this doc we'll be referring to xyz as request/response objects", because the term request/response may be confusing.  Everything is a request/response in a way.  Or maybe just "non-state" apis?

howto/serialize/README.md
```diff
@@ -0,0 +1,64 @@
+# Serialization in Dapr's SDKs
+
+An SDK for Dapr should provide serialization for two use cases. First, for API objects sent through request and response payloads. Second, for objects to be persisted. For both these use cases, a default serialization must be provided. In the Java SDK, it is the [DefaultObjectSerializer](https://dapr.github.io/java-sdk/io/dapr/serializer/DefaultObjectSerializer.html) class, providing JSON serialization.
```
LMWF: > a default serialization must be provided

-> "a default serialization is provided".

howto/README.md
artursouza: "Serialize objects" should be enough in this page. No need to be detailed since it requires further explanation.

howto/serialize/README.md
amanbha: There are following things and we should talk about serialization/desalinization for all:
1. State Management,
2. Service Invocation Request/Response.
3. Publishing Events.
4. Bindings.

For Actors:
1. Actor Proxy to Actor Service Communication Request/Response.
2. Actor state management.

artursouza: Done.

howto/serialize/README.md
```diff
@@ -0,0 +1,138 @@
...
+    DaprClient client = (new DaprClientBuilder()).build();
+    client.invokeBinding("sample", "My Message").block();
+```
+
+* Input binding:
+```java
+  @PostMapping(path = "/sample")
+  public void handleInputBinding(@RequestBody(required = false) byte[] body) {
+      String message = (new DefaultObjectSerializer()).deserialize(body, String.class);
+      System.out.println(message);
```
LMWF: It's helpful to the user if all the output of all "prints" are shown like in https://gobyexample.com/hello-world

howto/serialize/README.md
```diff
@@ -0,0 +1,138 @@
...
+  }
+```
+
+## Actor's method invocation
```
LMWF: Actor Method Invocation

artursouza: done.

PR #367, opened 2020-02-22
Document DAPR_HOST_IP in troubleshooting section

yaron2:
Part of https://github.com/dapr/dapr/issues/1087.

### DIFF
--- a/best-practices/troubleshooting/common_issues.md
+++ b/best-practices/troubleshooting/common_issues.md
@@ -141,3 +141,17 @@ kubectl logs <pod-name> <name-of-your-container>
 ```
 
 If running in Standalone mode, you should see the stderr and stdout outputs from your app displayed in the main console session.
+
+### I'm getting timeout/connection errors when running Actors locally
+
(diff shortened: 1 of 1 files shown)


best-practices/troubleshooting/common_issues.md
```diff
@@ -141,3 +141,17 @@ kubectl logs <pod-name> <name-of-your-container>
 ```
 
 If running in Standalone mode, you should see the stderr and stdout outputs from your app displayed in the main console session.
+
+### I'm getting timeout/connection errors when running Actors locally
+
+Each Dapr instance reports it's host address to the placement service. The placement service then distributes a table of nodes and their addresses to all Dapr instances. If that host address is unreachable, you are likely to encounter socket timeout errors or other variants of failing request errors.
+
+Unless the host name has been specified by setting an environment variable named `DAPR_HOST_IP` to a reachable, pingable address, Dapr will loop over the network interfaces and select the first non-loopback address it finds.
```
artursouza: mention that it is called HOST_IP in versions 0.4 and prior.

PR #381, opened 2020-03-03
Update docs structure and content

AaronCrawfis:
# Description

Various updates to the docs:
- Standard naming convention for all concepts: README.md under the concept folder
- Move api docs into the respective concept folder and link reference to the api docs
- Move redis configuration instructions into howtos
- Remove quickstart section (redundant with getting started)


AaronCrawfis:
Oh didn't know about the capitalization guidelines, I'll definitely keep it lower going forward.

yaron2:
Just to make sure, we did not move API reference under concepts in the end, right?

msfussell:
Correct. API reference should remain where it is. It did not move.

AaronCrawfis:
@yaron2 @msfussell Actually in this branch I moved them into concepts.

I still want to make the argument for having both the overview and API reference under concepts. Currently from the dev experience if you want to use a building block: from reading about it for the first time, understanding how it works, and reading the API reference you'll need to look in 2-3 places. When I got started writing demos this was very confusing for me and why I opened #85. For me it makes sense to have the level 100 (what is it), level 200 (how do I use it) and level 300 (how does it work) all in one place. Then under references you can link to all the API docs, just like you do with the CLI docs. 

Additionally this makes maintaining updates to the docs easier, so as a building block gets created or updated it will be one place to update content instead of 2.

But if you feel strongly the other way I'll revert back to the current structure.

msfussell:
Let' keep the API under the reference for now, since I feel there needs to be one place to find this reference.

AaronCrawfis:
@msfussell I agree there needs to be a reference, but I see that more as a central appendix with links into all the applicable documents.

I'll update the api docs to go back under reference for now, but I'd like to continue this discussion and make the argument to centralize building block documentation all in one location.

PR #404, opened 2020-03-09
Add components authorization How To

### LINKED ISSUE #391: Add How-To for component authorization

yaron2:
Closes #391

### DIFF
--- a/concepts/README.md
+++ b/concepts/README.md
@@ -32,7 +32,7 @@ The following are the building blocks provided by Dapr:
 | [**Distributed Tracing**](./distributed-tracing/README.md) | `TBD` |  Distributed tracing collects and aggregates trace events, metrics and performance numbers between Dapr instances. It allows you to trace the entire call chain across multiple services, or see call metrics on a user service. Dapr currently integrates with  [Open Census](https://opencensus.io/) and when ready [OpenTelemetry](https://opentelemetry.io/) for distributed tracing and metrics collection.
 | [**Actors**](./actors/README.md) | `/v1.0/actors` |  An actor is an isolated, independent unit of compute and state with single-threaded execution. Dapr provides an actor implementation based on the Virtual Actor pattern which provides a single-threaded programming model and where actors are garbage collected when not in use. See * [Actor Overview](./actor/actor_overview.md)
 
-## s
+## Components
  
 Dapr uses a modular design where functionality is delivered as a component. Each component has an interface definition.  All of the components are pluggable so that you can swap out one component with the same interface for another. The [components contrib repo](https://github.com/dapr/components-contrib) is where you can contribute implementations for the component interfaces and extends Dapr's capabilities.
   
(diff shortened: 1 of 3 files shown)


concepts/README.md
```diff
@@ -32,7 +32,7 @@ The following are the building blocks provided by Dapr:
 | [**Distributed Tracing**](./distributed-tracing/README.md) | `TBD` |  Distributed tracing collects and aggregates trace events, metrics and performance numbers between Dapr instances. It allows you to trace the entire call chain across multiple services, or see call metrics on a user service. Dapr currently integrates with  [Open Census](https://opencensus.io/) and when ready [OpenTelemetry](https://opentelemetry.io/) for distributed tracing and metrics collection.
 | [**Actors**](./actors/README.md) | `/v1.0/actors` |  An actor is an isolated, independent unit of compute and state with single-threaded execution. Dapr provides an actor implementation based on the Virtual Actor pattern which provides a single-threaded programming model and where actors are garbage collected when not in use. See * [Actor Overview](./actor/actor_overview.md)
 
-## Components
+## s
```
msfussell: Why this delete!!

yaron2: Fixed, by mistake

yaron2: Refresh.. fixed now

howto/README.md
```diff
@@ -43,6 +43,10 @@ Here you'll find a list of How To guides that walk you through accomplishing spe
 * [Configure component secrets using Dapr secret stores](./setup-secret-store)
```
msfussell: Whilst you are updating this readme.md can you add the topic https://github.com/dapr/docs/tree/master/howto/configure-redis  which was moved by Aaron, but no link added in the readme.md. 
I suggest that you put a entry under **both** State Management heading and Pub/Sub heading called "Setup Redis as a state store" and  ""Setup Redis for pub/sub"

yaron2: It'll be a separate PR, we should really keep them separate.

yaron2: I'll do it right after this is merged.

howto/components-scopes/README.md
```diff
@@ -0,0 +1,66 @@
...
+  - name: redisHost
+    value: redis-master:6379
+```
+
+In this example, the Redis component is only accessible to Dapr instances running inside the `production` namespace.
+
+### Example of component namsespacing in self-hosted mode
+
+In self hosted mode, a developer can specify the namespace to a Dapr instance by setting the `NAMESPACE` environment variable.
+If the `NAMESPACE` environment variable is set, Dapr will not load any component that does not specify the same namespace in its metadata.
```
msfussell: How is this set and is this a general name? Should this not be called "DAPR-NAMESPACE "?

yaron2: It's a general name, Kubernetes and a few other projects use it without prefixes. In the future if people have different wants, we can look into changing that.

yaron2: The namespace env var btw is not something new, we had it since 0.1.0

howto/components-scopes/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+Considering the example above, to tell Dapr which namespace it is deployed to, set the environment variable:
+
+```bash
+export NAMESPACE=production
```
msfussell: Add setx NAMESPACE "Production" as Windows example

yaron2: Done

PR #415, opened 2020-03-10
Update service_invocation_api.md

yaron2:
Remove `200 OK` as a successful response for service invocation, since the status code is returned by the callee.


yaron2:
> To be clear, I think what you have in the description of this PR should be in the spec.

You're right, on it. Check now.

PR #440, opened 2020-03-13
Improve appinsight distributed trace setup guide and add deployment yamls

youngbupark:
# Description

Add local forwarder docker and deployment yamls.

* Pushed localforwarder docker image to daprio org : https://hub.docker.com/r/daprio/dapr-localforwarder/tags
* Updated docs

## Issue reference

https://github.com/dapr/components-contrib/issues/207

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/howto/diagnose-with-tracing/azure-monitor.md
+++ b/howto/diagnose-with-tracing/azure-monitor.md
@@ -2,6 +2,8 @@
 
 Dapr integrates with Application Insights through OpenTelemetry's default exporter along with a dedicated agent known as [Local Forwarder](https://docs.microsoft.com/en-us/azure/azure-monitor/app/opencensus-local-forwarder).
 
+> Note: Local forwarder is still under preview, but being deprecated. Application insights team recommends to use [Opentelemetry collector](https://github.com/open-telemetry/opentelemetry-collector) (which is in alpha state) for the future so that we're working on moving from local forwarder to [Opentelemetry collector](https://github.com/open-telemetry/opentelemetry-collector).
+
 ## How to configure distributed tracing wit Application insights
 
 The following steps will show you how to configure Dapr to send distributed tracing data to Application insights.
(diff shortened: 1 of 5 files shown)


howto/diagnose-with-tracing/azure-monitor.md
```diff
@@ -1,28 +1,55 @@
...
+3. Get Application insights Intrumentation key from your application insights page
+4. Go to `Configure -> API Access`
+5. Click `Create API Key`
+6. Select all checkboxes under `Choose what this API key will allow apps to do:`
+   - Read telemetry
+   - Write annotations
+   - Authenticate SDK control channel
+7. Generate Key and get API key
 
 ### Setup the Local Forwarder
```
AaronCrawfis: @youngbupark we should have a note here mentioning that the dapr local forwarder image is for testing purposes only, and we do not support it in production scenarios. We should point to the dockerfile and say teams can use this to create their own image

youngbupark: I added some notes.I did not use word `testing` purpose. Instead, I used the word `preview`.  but feel free to edit the file directly if you want :)

AaronCrawfis: Hmmm not sure if we should use the term `preview` if we don't have eventual plans to turn it into a supported feature. Don't want companies to take dependencies thinking we'll continue to develop it. But you're right, `testing` might also not be great.

@msfussell what are your thoughts?

youngbupark:
cc/ @RicardoNiepel

howto/diagnose-with-tracing/azure-monitor.md
amanbha: I believe the term preview is used here for LocalForawrder which is from ApplciationInsights team and thus preview is used to refer to it.

Issue #416, opened 2020-03-10
Add Error Codes returned by Dapr apis to reference


amanbha:
Add Error Codes returned by Dapr apis to reference under https://github.com/dapr/docs/tree/master/reference

PR #444, opened 2020-03-17
Adding error codes returned by Dapr http calls.

amanbha:
Closes https://github.com/dapr/docs/issues/416


reference/api/error_codes.md
```diff
@@ -0,0 +1,38 @@
...
+ERR_ACTOR_REMINDER_GET | Error getting a reminder for an actor.
+ERR_ACTOR_INVOKE_METHOD | Error invoking a method on an actor.
+ERR_ACTOR_STATE_DELETE | Error deleting the state for an actor.
+ERR_ACTOR_STATE_GET | Error getting the state for an actor.
+ERR_ACTOR_STATE_TRANSACTION_SAVE | Error storing actor state transactionally.
+ERR_PUBSUB_NOT_FOUND | Error referencing the Pub/Sub component in Dapr runtime.
+ERR_PUBSUB_PUBLISH_MESSAGE  | Error publishing a message.
+ERR_PUBSUB_CLOUD_EVENTS_SER | Error serializing Pub/Sub event envelope.
+ERR_STATE_STORE_NOT_FOUND | Error referencing a state store not found.
+ERR_STATE_GET | Error getting a state for state store.
```
LMWF: "a state" -> "state"

LMWF: Same for next 2

reference/api/error_codes.md
```diff
@@ -0,0 +1,38 @@
...
+ERR_STATE_DELETE | Error deleting a state from state store.
+ERR_STATE_SAVE | Error saving a state in state store.
+ERR_INVOKE_OUTPUT_BINDING | Error invoking an output binding.
+ERR_MALFORMED_REQUEST | Error with a malformed request.
+ERR_DIRECT_INVOKE | Error in direct invocation.
```
LMWF: I believe this only happens in a message from daprd to app.  Can the description say that?  Otherwise "direct invocation" means nothing to the user.

reference/api/error_codes.md
```diff
@@ -0,0 +1,38 @@
...
+    "errorCode": "ERR_STATE_GET",
+    "message": "Requested state key does not exist in state store."
+}
+```
+
+Following table lists the error codes returned by Dapr runtime:
+
+Error Code | Description
+--------- | -----------
+ERR_ACTOR_INSTANCE_MISSING | Error getting an actor instance. This means that actor is now hosted in some other service replica.
```
LMWF: Doesn't have to be here, but we should publish which ones are retryable vs not.  For example the this first one is.

amanbha: @LMWF  thats a good point, we can add them to this list in future.

PR #457, opened 2020-03-25
Add the Dapr /healthz API spec

### LINKED ISSUE #455: Add the new Dapr healthz api specification doc
As the new Dapr /healthz API is added, this issue is to add the API spec document.

Feature Reference : 
https://github.com/dapr/dapr/issues/1288

shalabhms:
# Description

As the new Dapr /healthz API is added, this issue is to add the API spec document.

Feature Reference :
dapr/dapr#1288

## Issue reference

Closes #455 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/reference/api/health_api.md
+++ b/reference/api/health_api.md
@@ -29,6 +29,5 @@ daprPort | The Dapr port.
 
 ```shell
 curl http://localhost:3500/v1.0/healthz
-  -H "Content-Type: application/json"
 ```
 
--- a/reference/api/health_api.md
+++ b/reference/api/health_api.md
@@ -1,0 +1,34 @@
+# Dapr health API reference
+
+Dapr provides health checking probes that can be used as readiness or liveness of Dapr.
+
+### Get Dapr health state
+
+Gets the health state for Dapr.
+
+#### HTTP Request
+
+```http
+GET http://localhost:<daprPort>/v1.0/healthz
+```
+
+#### HTTP Response Codes
+
+Code | Description
+---- | -----------
+200  | health is ok
+500  | health is not ok
+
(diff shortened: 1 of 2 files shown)


reference/api/health_api.md
```diff
@@ -0,0 +1,34 @@
...
+Parameter | Description
+--------- | -----------
+daprPort | The Dapr port.
+
+#### Examples
+
+```shell
+curl http://localhost:3500/v1.0/healthz
+  -H "Content-Type: application/json"
+```
```
youngbupark: 200 OK is supposed to have request body. I wonder if we need to return simple body. e.g.

```
{"status":"ok"}
```

@yaron2 @amanbha how do you think ?

yaron2: We need to remove the header with the content type.  We don't have to return a response, a status code with no body is valid, this is what Kubernetes health checks expect also. They only examine the status code.

shalabhms: Thanks, removed the header with the content type.

reference/api/health_api.md
```diff
@@ -0,0 +1,33 @@
...
+
+```http
+GET http://localhost:<daprPort>/v1.0/healthz
+```
+
+#### HTTP Response Codes
+
+Code | Description
+---- | -----------
+200  | health is ok
```
yaron2: I would change to `dapr is healthy` and `dapr is not healthy`

yaron2: Other than that, LGTM

shalabhms: Done as suggested. thanks.

PR #468, opened 2020-03-29
Improve example creating actor

### LINKED ISSUE #467: Improve FAQ example about actors
The FAQ chapter [Differences between Dapr from an actor framework](https://github.com/dapr/docs/blob/master/FAQ.md#differences-between-dapr-from-an-actor-framework) has an example that is not making much sense to me.

### Current text

> Creating a new actor follows a local call like http://localhost:3500/v1.0/actors/<actorType>/<actorId>/meth... for example http://localhost:3500/v1.0/actors/myactor/50/method/getData to call the getData method on myactor with id=50

### Issues

1. It starts about creating a new actor, but the URL is resembling more calling a method.
1. The placeholders in the URL are not escaped so they don't end up in the rendered page.
1. The example describes calling a method on an instance of an actor, not just creating an actor.

### Solutions

It's not clear what the original intention was behind this example, as it is already this way in the original [wiki](https://github.com/dapr/dapr/wiki/FAQ). 
I see 3 possible solutions to fix this paragraph.

1. Change first sentence to make it about calling a method, escape placeholders and fix URL.
1. Escape placeholders, fix URL and change sample to only create an instance, remove mention of method.
1. Escape placeholders, fix URL and add sample about creating an instance of an actor. Add additional text and URI about calling a method on an actor fitting the existing example.

My suggestion is to pick the first one, as it's enough to prove the platform agnostic possibility to call actors. If one would need more details to work with actors, the FAQ is not the place to be.

If there is consensus, I like to put a PR in.

eNeRGy164:
# Description

* Escaped placeholders
* Small textual changes,  f.e. to underscore the instantiation of the actor. 

## Issue reference

Closes #467

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -37,7 +37,7 @@ Virtual actors capabilities are one of the building blocks that Dapr provides in
 
 Creating a new actor follows a local call like http://localhost:3500/v1.0/actors/\<actorType>/\<actorId>/…, for example <http://localhost:3500/v1.0/actors/myactor/50/method/getData> to call the `getData` method on the newly created `myactor` with id `50`.
 
-The Dapr runtime SDKs have language specific actor frameworks. The .NET SDK for example has C# actors. You will see all the SDKs have an actor framework that fits with the language.
+The Dapr runtime SDKs have language specific actor frameworks. The .NET SDK for example has C# actors. The goal is for all the Dapr language SDKs to have an actor framework.
 
 ## Developer language SDKs and frameworks
 
--- a/FAQ.md
+++ b/FAQ.md
@@ -35,10 +35,9 @@ Also Dapr is about more than just actors. It provides you with a set of best pra
 
 Virtual actors capabilities are one of the building blocks that Dapr provides in its runtime. With Dapr because it is programming language agnostic with an http/gRPC API, the actors can be called from any language. This allows actors written in one language to invoke actors written in a different language.
 
-Creating a new actor follows a local call like http://localhost:3500/v1.0/actors/<actorType>/<actorId>/meth...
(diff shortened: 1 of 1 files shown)


FAQ.md
```diff
@@ -35,10 +35,9 @@ Also Dapr is about more than just actors. It provides you with a set of best pra
 
 Virtual actors capabilities are one of the building blocks that Dapr provides in its runtime. With Dapr because it is programming language agnostic with an http/gRPC API, the actors can be called from any language. This allows actors written in one language to invoke actors written in a different language.
 
-Creating a new actor follows a local call like http://localhost:3500/v1.0/actors/<actorType>/<actorId>/meth...
-for example http://localhost:3500/v1.0/actors/myactor/50/method/getData to call the getData method on myactor with id=50
+Creating a new actor follows a local call like http://localhost:3500/v1.0/actors/\<actorType>/\<actorId>/…, for example <http://localhost:3500/v1.0/actors/myactor/50/method/getData> to call the `getData` method on the newly created `myactor` with id `50`.
 
-The Dapr runtime SDKs have language specific actor frameworks. The the .NET SDK for example has C# actors. You will see all the SDKs have an actor framework that fits with the language
+The Dapr runtime SDKs have language specific actor frameworks. The the .NET SDK for example has C# actors. You will see all the SDKs have an actor framework that fits with the language.
```
msfussell: Noticed there is a typo in this line, which would also be good to fix "The the .Net" should be "The .NET SDK"

msfussell: This sentence is also a strangely worded "You will see all the SDKs have an actor framework that fits with the language".
 I would update this to simply "The goal is for all the Dapr language SDKs to have an actor framework."

Thanks for doing these updates. Appreciate this attention to detail.

Issue #214, opened 2019-10-31
How to glue Input and Output bindings?


abhirockzz:
## Ask your question here

In the [Dapr Azure Friday episode](https://youtu.be/mPVnu4W0xzQ?t=150),  @yaron2 showed a demo that uses Input and Output bindings in a single Dapr app using a specific format of the response (pretty cool!)

Questions/comments:

1. I looked up the samples and how-to but did not find a reference to such an example
2. Is there documentation on the same (in bindings section)?
3. Are there any other patterns/techniques which can be used to "glue" components together? e.g. is it possible to get an event via Input binding and push it to a state store?
4. Also, can I get a reference to the part of code (in Dapr runtime I am assuming) which accomplishes this (sorry for being lazy!)

I found this to be really powerful and it would be great to capture this (and other patterns) in the docs

yaron2:
Hi,

We have an [input bindings How To](https://github.com/dapr/docs/tree/master/howto/trigger-app-with-input-binding) and [output bindings How To](https://github.com/dapr/docs/tree/master/howto/send-events-with-output-bindings).

It should be fairly easy to connect the two - just one app that listens on an endpoint (for input bindings) and that sends an HTTP message to invoke an output binding.

There's also a sample here for bindings:
https://github.com/dapr/samples/tree/master/5.bindings

Let us know if there's any more specific help you think is needed.

abhirockzz:
Thanks @yaron2 . Yes I did check the how-tos as well as the sample. What I was referring to is this technique (which you showcased in the Azure Friday demo) to glue the bindings together in one app

```
...
app.post('/eventhubs', (req, resp) =>){
  console.log(req.body)
  res.json({to: ['storage'], data: req.body})
})
...
```
If I understand correctly, this means

- get the message using `eventhubs` input binding
- send it to `storage` output binding (which in this example was Azure Blob)

yaron2:
Oh that!

You understand correctly.

Yeah that's a neat trick we have that's not documented at all right now. We'll work on that.

You can dispatch the same message to multiple outputs, with either ordered delivery or you can fan out in parallel by adding another property: 

`concurrency: parallel`

yaron2:
Oh and to your question in .3, yes it's possible.. you can also send the data to a store using a response format.

abhirockzz:
Exactly! It was really cool and would be great to have documentation and the input/output combinations with which this is possible (maybe all of them)

RicardoNiepel:
[...]

PR #471, opened 2020-03-30
Add optional Binding HTTP response body into spec

RicardoNiepel:
# Description

If an input binding is invoked, the applicaiton can return a HTTP response body to let Dapr automatically send data to a state store or to a output binding.

If multiple output bindings are specified, it is possible to configure if the data is send in parallel or sequential. 

## Issue reference

Please reference the issue this PR will close: #214

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


RicardoNiepel:
@yaron2 pls review

reference/api/bindings_api.md
```diff
@@ -80,6 +80,73 @@ def incoming():
     return "Kafka Event Processed!"
 ```
 
+### Binding subscription
+
+On startup Dapr will invoke the following endpoint on user code to discover binding subscriptions:
```
amanbha: This needs rewording: Bindings are discovered from component yamls. Dapr calls this endpoint on startup to ensure that app indeed is handling it, if the app doesnt have the endpoint, dapr ignores it.

reference/api/bindings_api.md
```diff
@@ -80,6 +80,73 @@ def incoming():
     return "Kafka Event Processed!"
 ```
 
+### Binding subscription
```
amanbha: Binding subscription doesn't seem the correct term here. It has potential to confuse customers with pubsub.

msfussell: I addressed the comments from Aman. Take a look

Issue #529, opened 2020-04-20
Storage Queue Binding Documentation


WillEastbury:
## Storage Queue Binding does not deploy as documented to K8S, but will work locally

ttlInSeconds parameter needs to be doublequoted or you get a very, very arcane error message in the dapr sidecar logs complaining about an error from the ReadString method requires \" or n which took me quite a long time to figure out.

This only seems to not work on K8s

PR #530, opened 2020-04-20
Wrap value in doublequotes for ttlInSeconds value.

WillEastbury:
# Description

ttlInSeconds parameter needs to be doublequoted or you get a very, very arcane error message complaining about an error from the ReadString method requires \" or n which took me quite a long time to figure out.

## Issue reference

Please reference the issue this PR will close: #529 

https://github.com/dapr/docs/issues/529

## Checklist

[X] Extended the documentation


yaron2:
Thanks for this.

Yes, this is due to the way Kubernetes treats string maps on objects.

PR #542, opened 2020-04-25
Updated configuration docs

msfussell:
Updated the Configuration concepts doc and this then made me update several other files related to configuration


concepts/configuration/README.md
```diff
@@ -1,8 +1,140 @@
...
+- [Kubernetes sidecar configuration](#Kubernetes-sidecar-configuration)
+- [Sidecar Configuration settings](#sidecar-configuration-settings)
+- [Kubernetes control plane configuration](#Kubernetes-control-plane-configuration)
+- [Control plane configuration settings](#control-plane-configuration-settings)
 
-* [distributed tracing](../observability/traces.md)
-* [custom pipeline](../middleware/README.md)
+## Self hosted sidecar configuration  
+In self hosted mode the Dapr configuration is a configuration file, for example `myappconfig.yaml`. By default Dapr side looks in the `components/` sub-folder under the folder where you run your application for a configuration file.
+
```
orizohar, line 14: Would be good to also mention that when running via the CLI, it will create the components directory with default configurations if the directory is not found

PR #544, opened 2020-04-26
create deployment vs run

mchmarny:
Switching to `create deployment` as the `kubectl run` no longer creates deployment in the current format. Rest of the instructions are fine.

`kubectl run` to `kubectl create deployment`


mchmarny:
On k8s 1.15 the `run` command doesn’t seem to create deployment anymore so more explicit `create deployment` change

PR #546, opened 2020-04-27
Documenting updates regarding express.js > 4.16 and the expected cont…

raykao:
…ent-type

In express.js > 4.16 the bodyparser has been added back in to express core.  As such you will receive a deprecation warning regarding the use of:

```
var bodyParser = require(body-parser);
// ...
app.use(bodyParser.json());
// ...
```

Which has been replaced simple with:

```
app.use(express.json();
```

You will also need to modify the accepted content-types options object (```{types: []}```) passed into ```express.json()``` to allow/accept the ```application/cloudevents+json``` content-type, since Dapr leverages/uses/conforms to the [CloudeEvents Spec](https://cloudevents.io).

```
app.use(express.json({type: ["application/json", "application/cloudevents+json"]}));
```

Lastly there may/may not be a need to change the ```req.body``` object to reflect **only** the message data payload based on the above ```content-type: application/cloudevents+json``` as opposed to the entire CloudeEvent spec:

```
app.use(function(req, res, next){
    if(req.headers["content-type"] == "application/cloudevents+json") {
        req.body = req.body.data
    }
    next();
});
```

# Description

_Please explain the changes you've made_

## Issue reference

Please reference the issue this PR will close: #_[issue number]_

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


howto/consume-topic/README.md
```diff
@@ -34,7 +34,7 @@ To deploy this into a Kubernetes cluster, fill in the `metadata` connection deta
 To subscribe to topics, start a web server in the programming language of your choice and listen on the following `GET` endpoint: `/dapr/subscribe`.
 The Dapr instance will call into your app, and expect a JSON value of an array of topics.
 
-*Note: The following example is written in node, but can be in any programming language*
+*Note: The following example is written in node (express.js < 4.16), but can be in any programming language*
```
amanbha, line 37: Can we just update it for >4.16, it would simplify it.

howto/consume-topic/README.md
```diff
@@ -68,6 +86,8 @@ app.post('/topic1', (req, res) => {
 })
 ```
 
+NOTE: Express
```
orizohar, line 89: This looks like an unintended addition. Correct?

howto/consume-topic/README.md
orizohar: Agree. Lets have the code we show be the latest and greatest version of express.

orizohar:
@raykao - can you please address the PR change requests?

msfussell:
@raykao - Would you be able to address the comments here?
Also note that subscribe now takes pubsubName like this

app.get('/dapr/subscribe', (_req, res) => {
    res.json([
        {
            pubsubName: "pubsub",
            topic: "A",
            route: "A"
        },
        {
            pubsubName: "pubsub",
            topic: "B",
            route: "B"
        }
    ]);
});

howto/consume-topic/README.md
```diff
@@ -63,6 +63,24 @@ app.post('/orders', (req, res) => {
 app.listen(port, () => console.log(`consumer app listening on port ${port}!`))
 </pre>
 
+**NOTE:** For node/express > 4.16 you will need to add the following:
```
artursouza, line 66: These instructions should not be part of the example. Also, they might become stale in a future release on express (5.0, for example). We should stick to one version and just document it. Please, just rewrite this note to call out the express version and that the code might not work with future express versions.

howto/consume-topic/README.md
artursouza: The wording in this phrase is confusing (IMO). Just add a note on the express version below and it should be enough.

I agree that we should track the work to upgrade all node examples to the latest version of express.

artursouza (changes requested):

AaronCrawfis:
Closing this PR, as we're moving docs to docs.dapr.io shortly and we've updated the guide to follow our quickstart and only have Node > 4.16. 

https://docs.dapr.io/developing-applications/building-blocks/pubsub/howto-publish-subscribe/

Thank you for your contribution! Please PR any more updates/clarifications you find over at the new docs.dapr.io

PR #575, opened 2020-05-09
Twitter input binding

mchmarny:
List Twitter input binding on the list of bindings


mchmarny:
It is input only, md in browser edit hard. fixed

PR #578, opened 2020-05-14
Howto for extending Dapr using custom components

pulkitmehra:
Co-authored-by: Phil Kedy <phil.kedy@gmail.com>

# Description

Organization may have a requirement to add custom functionality to Dapr, such as integration with propriety or legacy systems. Custom Components is a point of extension that allows third-party functionality to run alongside the existing building blocks that ship with Dapr. For example, you have an existing library for encryption and decryption that you would like to expose as a building block to your applications. Applications will connect to this custom component via a gRPC endpoint which is the same mechanism the Dapr SDKs uses.

In this feature we'll be extended Dapr by using it as a library. Simply create a new Go project and copy the original [main.go](https://github.com/dapr/dapr/blob/master/cmd/daprd/main.go) from Dapr Github source code to this new project(using Dapr as library here), next stitch together all the Components(Dapr runtime) in the same way as it was present in the original `main.go`, next register Custom component in runtime. The new binary created after compilation is a custom `Daprd` binary with a new gRPC endpoint as a custom component.

## Issue reference

Please reference the issue this PR will close: #_[issue number]_

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


pulkitmehra:
This depends upon following PRs
1. [Component-Contrib](https://github.com/dapr/components-contrib/pull/335)
2. [Dapr](https://github.com/dapr/dapr/pull/1561)
3. [Samples](https://github.com/dapr/samples/pull/209)

mchmarny:
Given that this will depend on go anyway. Should we consider building the client component portion of this into the go SDK? We already have the [gRPC serving](https://github.com/dapr/go-sdk/blob/main/service/grpc/Readme.md) capability there so adding this would be trivial.

AaronCrawfis:
Hi @pulkitmehra, we've reviewed this design and decided to tackle custom components differently in a future Dapr release. Thank you for your contributions to Dapr, we greatly value your expertise and look forward to future input & contributions. Please let us know if you have any questions!

ewassef:
Hello @AaronCrawfis . Do you have any updates on when / how this would be implemented? We have the same requirement and the inability to extend or add components has been a roadblock to the full adoption of Dapr. Namely, we want to do some payload processing of the messages for pub-sub for GDPR purposes and allow for tokenization. 
I think we can do some funky other sidecars with a tiny Redis endpoint and then our custom code as a consumer then forward it elsewhere, but this is not a good solution

PR #606, opened 2020-05-26
Added contribution guidance links in PR template

orizohar:
Helping contributors to docs repo make high quality contributions by adding links to the contribution guidance in the PR template.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,3 +1,7 @@
+Thank you for helping make the Dapr documentaion better!
+
+Please note all documentation contributions must be in accordance with the [official contribution guidance](https://github.com/dapr/docs/blob/master/contributing/README.md).
+
 # Description
 
 _Please explain the changes you've made_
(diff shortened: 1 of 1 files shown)


mchmarny:
`must be in accordance with the official contribution guidance` seems little strong for an OSS project IMO

orizohar:
> `must be in accordance with the official contribution guidance` seems little strong for an OSS project IMO

Can you suggest a different phrasing that you think is more appropriate? 
I think we want to encourage people to follow this. Recently I'm coming across PRs that have very different structure and style and I think we want to try and create consistency where ever possible.

mchmarny:
Always best if the template itself is self-explanatory to the point that makes following the right format easy. We shouldn't also try to prevent any possibility of "different structure and style" and rely rather on PR reviewers to ensure the right content and ask for clarification, additional content or change to styles if necessary. Suspect the minimal valuable content will be different for different types of PRs

PR #658, opened 2020-06-24
Update default config to daprsystem config

youngbupark:
## Description

Update default config to daprsystem config.

## Issue reference

https://github.com/dapr/dapr/issues/1408


youngbupark:
@yaron2 are there additional docs we need to update?

yaron2:
> @yaron2 are there additional docs we need to update?

Not that I'm aware of.

PR #660, opened 2020-06-25
issue #1698 resolution

KaiWalter:
## Description

Added a note explaining the resolution for https://github.com/dapr/dapr/issues/1698 - that a complete Application Insights application map is only provided, when all apps and services invoking method are operated with a Dapr sidecar.

## Issue reference

_Please reference the issue this PR will close: https://github.com/dapr/dapr/issues/1698_


howto/diagnose-with-tracing/azure-monitor.md
```diff
@@ -162,6 +162,8 @@ Deploy and run some applications. After a few minutes, you should see tracing lo
 
 ![Application map](../../images/azure-monitor.png)
 
+> **NOTE**: For **Application Map** to display the correct topology - and with that rendering `requests` and `dependencies` item types correctly - all apps or services which invoke methods over Dapr need to be operated with a Dapr sidecar. Dapr client invocations from an outside app or service (not operated with a sidecar), are not captured.
+
 ## Tracing configuration
```
mchmarny, line 167: This makes sense and is a good clarification. Propose rewording though to make it more clear that only direct invocations not going through Dapr API (and without a valid trace ID) will not be shown in topology:

> **NOTE:** Only operations going through Dapr API exposed by Dapr sidecar (e.g. service invocation or event publishing) will be displayed in Application Map topology. Direct service invocations (not going through the Dapr API) will not be shown.

KaiWalter: thanks for the suggestion which definitely is more clear

Issue #667, opened 2020-06-30
Update Cosmos DB howto for actors


LMWF:
Update Cosmos documentation for changes in https://github.com/dapr/components-contrib/issues/152

PR #668, opened 2020-06-30
Update cosmos documentation for transaction changes

LMWF:
## Description

This updates the doc for Cosmos DB:
- modifies instructions on how to set the partition key in cosmos.  This is a breaking change.
- describes partition key usage for actors, modified the section on this for non-actors
- adds an example yaml

The 2 corresponding PRs in the other repos must go in first.
https://github.com/dapr/components-contrib/pull/372
https://github.com/dapr/dapr/pull/1724

## Issue reference

Please reference the issue this PR will close: #667


howto/setup-state-store/setup-azure-cosmosdb.md
```diff
@@ -4,7 +4,7 @@
 
 [Follow the instructions](https://docs.microsoft.com/en-us/azure/cosmos-db/how-to-manage-database-account) from the Azure documentation on how to create an Azure CosmosDB account.  The database and collection must be created in CosmosDB before Dapr consumes it.  
 
-**Note : The partition key for the collection must be "/id".**
+**Note : The partition key for the collection must be "/partitionKey".  This is case-sensitive.**
```
mchmarny, line 7: The way this is phrased right now it sounds like the actual value of the partition key "must be" `/partitionKey` which is not the case, right? Rewrite to make it clear that the case of the partitioning key must match the case of the document's property.

howto/setup-state-store/setup-azure-cosmosdb.md
```diff
@@ -63,6 +63,33 @@ spec:
...
+    value: c1
+```
+
+If you wish to use this as the actor secret store, append the following to the yaml.
+
+```yaml
+  - name: actorStateStore
+    value: "true"
+```
+
```
mchmarny, line 92: What is the purpose of this new YAML example? The current document already outlines the plane YAML approach where the values include placeholder (e.g. `<REPLACE-WITH-URL>`) and a Kubernetes example which includes value from secret approach:

```yaml
secretKeyRef:
      name: <KUBERNETES-SECRET-NAME>
      key: <KUBERNETES-SECRET-KEY>
```

LMWF: In my opinion sometimes a template with "[abc]" but without actual values is unclear, and having real values (or values following the format of real values) helps clarify.  Not a strong opinion though, if this looks weird I'll remove.

howto/setup-state-store/setup-azure-cosmosdb.md
LMWF: It does have to be exactly `/partitionKey`.

howto/setup-state-store/setup-azure-cosmosdb.md
mchmarny: No, nothing weird, just I haven't seen actual values populates into docs for other bindings. It's possible I didn't look far enough. @orizohar thoughts?

howto/setup-state-store/setup-azure-cosmosdb.md
LMWF: To elaborate, in general the field doesn't have to have this name, "id" and the correspoding "/id" are fine.  But this is a ripple from the requirement to support multi-item transactions.  In that context all items in the transaction must be in the same partition.  We were earlier using id for this.  But id is the key from the users view, so it must be unique.  So "partitionKey" was added as another field to represent the partition.  It's used in actor code to make sure items for the same actor end up in the same partition.

howto/setup-state-store/setup-azure-cosmosdb.md
```diff
@@ -75,13 +101,15 @@ kubectl apply -f cosmos.yaml
...
 ### Running locally
 
-To run locally, create a `components` dir containing the YAML file and provide the path to the `dapr run` command with the flag `--components-path`.
+To run locally, create a YAML file described above and provide the path to the `dapr run` command with the flag `--components-path`.  See [this](https://github.com/dapr/cli#use-non-default-components-path) or run `dapr run --help` for more information on the path.
 
 ## Partition keys
 
-The Azure CosmosDB state store will use the `key` property provided in the requests to the Dapr API to determine the partition key.
+For **non-actor** state operations, the Azure CosmosDB state store uses the `key` property provided in the requests to the Dapr API to determine the partition key.  This can be overridden by specifying a metadata field in the request with a key of `partitionKey` and a value of the desired partition.
+
```
msfussell, line 109: Can you provide an example of this "by specifying a metadata field in the request with a key of `partitionKey` and a value of the desired partition." to show how this metadata can be used

LMWF: Reorganized this section and added metadata with partitionKey examples.

PR #675, opened 2020-07-08
PostgreSQL docs

brooke-hamilton:
## Description

Documentation for the PostgreSQL state store.

## Issue reference

This PR applies to issue https://github.com/dapr/components-contrib/issues/225, and Pull Request https://github.com/dapr/components-contrib/pull/353


howto/setup-state-store/setup-postgresql.md
```diff
@@ -0,0 +1,56 @@
...
+```yaml
+apiVersion: dapr.io/v1alpha1
+kind: Component
+metadata:
+  name: statestore
+spec:
+  type: state.postgresql
+  metadata:
+  - name: connectionString
+    value: "<CONNECTION STRING>"
```
msfussell, line 38: Can you provide an example here? We have found that with an example, this helps people. Also **if** you can provide some guidance on how you get the connection string, that would help.

brooke-hamilton: I added an example and more guidance for how to create the connection string.

brooke-hamilton:
Fixed an upstream conflict

Issue #708, opened 2020-07-24
Changes after normalizing arguments across dapr cli, daprd and k8s annotations


pruthvidhodda:
## Describe the proposal

This issue tracks the changes needed in docs after [normalizing arguments across dapr cli, daprd and k8s annotations.](https://github.com/dapr/dapr/issues/1717)

[Parent Epic](https://github.com/dapr/dapr/issues/1717)

PR #737, opened 2020-08-12
Replace all deprecated cli, daprd options and k8s annotations

pruthvidhodda:
## Description

Replaced all deprecated cli, daprd options and k8s annotations

## Issue reference

_Please reference the issue this PR will close: #708


yaron2:
This should probably only be merged prior to the next release that will enable the changes here.

pruthvidhodda:
Our docs are versioned so this need not wait till the next release - unless we think it confuses the users.

PR #759, opened 2020-08-20
Removing mention of dapr init flag --install-path

orizohar:
This PR removes the mention of of --install-path flag from dapr init as the flag has been removed in v0.10.0


pruthvidhodda:
It's at two other places if I remember correctly.

PR #787, opened 2020-09-12
docs: add opa middleware docs

### LINKED ISSUE #786: Add Opa Middleware Docs
## Describe the proposal
Add basic document and examples for the new Open Policy Agent Middleware: https://github.com/dapr/dapr/issues/1617

RichiCoder1:
## Description

Adds a reference how-to for using the opa middleware.

## Issue reference

Closes #786

### DIFF
--- a/howto/policies-with-opa/README.md
+++ b/howto/policies-with-opa/README.md
@@ -85,6 +85,8 @@ type HTTPRequest struct {
   // The query broken down into keys and their values
   query map[string][]string
   // The request headers
+  // NOTE: By default, no headers are included. You must specify what headers
+  // you want to recieve via `spec.metadata.includedHeaders` (see above)
   headers map[string]string
   // The request scheme (e.g. http, https)
   scheme string
@@ -140,7 +142,21 @@ default allow = {
 }
 ```
 
+### Adding Request Headers
 
+You can also set additional headers to allowed ongoing request via the the same method:
+
+```go
+package http
+
+default allow = false
+
+allow = { "allow": true, "additional_headers": { "X-JWT-Payload": payload } } {
+  not input.path[0] == "forbidden"
+  # Where `jwt` is the result of another rule
+  payload := base64.encode(json.marshal(jwt.payload))
+}
+```
 
 
 ## Related links
--- a/howto/policies-with-opa/README.md
+++ b/howto/policies-with-opa/README.md
@@ -1,0 +1,58 @@
+# Apply Open Policy Agent Polices
+
+The Dapr Open Policy Agent (OPA) [middleware](https://github.com/dapr/docs/blob/master/concepts/middleware/README.md) allows applying [OPA Policies](https://www.openpolicyagent.org/docs/latest/http-api-authorization/) to Dapr requests. This can be used to apply reusable authorization policies to app endpoints. 
+
+## Middleware Component Definition
+```yaml
+apiVersion: dapr.io/v1alpha1
+kind: Component
+metadata:
+  name: my-policy
+  namespace: default
+spec:
+  type: middleware.http.opa
+  metadata:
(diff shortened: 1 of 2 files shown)


RichiCoder1:
Huh, just realized that component metadata needs to be in the key value format :<

RichiCoder1:
Updating the docs now, I wonder if I shouldn't drop the header restriction. It makes sense in Envoy since you're sending potentionally sensitive headers to an external service, but there's not good reason I can think of to exclude them by default here.

howto/policies-with-opa/README.md
```diff
@@ -0,0 +1,165 @@
...
+apiVersion: dapr.io/v1alpha1
+kind: Component
+metadata:
+  name: my-policy
+  namespace: default
+spec:
+  type: middleware.http.opa
+  metadata:
+    # `includedHeaders` is a comma-seperated set of case-insensitive headers to include in the request input.
+    # By default, request headers are not passed to the policy. To use an incoming header to make an authz decision, you must include it.
```
mchmarny, line 16: Consider: 
> Request headers are not passed to policy by default. Include to use incoming headers in authz decisions

howto/policies-with-opa/README.md
```diff
@@ -0,0 +1,165 @@
...
+    - name: includedHeaders
+      value: "x-my-custom-header, x-jwt-header"
+
+    # `defaultStatus` is the status code to return for denied responses if not specified in the `allow` result.
```
mchmarny, line 20: consider:  is the status code that will be returned with denied responses

howto/policies-with-opa/README.md
```diff
@@ -0,0 +1,165 @@
...
+
+default allow = {
+  "allow": false,
+  "status_code": 401
+}
+```
+
+### Adding Response Headers
+
+What if you you wanted to do a redirect instead? You can accomplish this using the ability to add headers to the returned result like so:
```
mchmarny, line 131: To redirect, add headers to the returned result:

howto/policies-with-opa/README.md
```diff
@@ -0,0 +1,165 @@
...
+  "status_code": 301,
+  "additional_headers": {
+    "Location": "https://my.redirect.site"
+  }
+}
+```
+
+### Adding Request Headers
+
+You can also set additional headers to allowed ongoing request via the the same method:
```
mchmarny, line 147: You can set additional headers using similar method:

howto/policies-with-opa/README.md
```diff
@@ -0,0 +1,165 @@
...
+  // you want to recieve via `spec.metadata.includedHeaders` (see above)
+  headers map[string]string
+  // The request scheme (e.g. http, https)
+  scheme string
+}
+```
+
+## Result
+
+The policy must set `data.http.allow` with either a `boolean` value, or an `object` value with an `allow` boolean property. A `true` `allow` will allow the request through, while a `false` value will reject the request with a `403` status. So for example, the following policy would return a `403 - Forbidden` for all requests:
```
mchmarny, line 98: The policy must set `data.http.allow` with either a `boolean` value, or an `object` value with an `allow` boolean property. A `true` `allow` will allow the request, while a `false` value will reject the request with a `403` status. The following policy demonstrates `403 - Forbidden` result for all requests:

mchmarny (changes requested):
Very detailed, thanks for putting this together. I've made a few clarifications but otherwise LGTM.

howto/policies-with-opa/README.md
RichiCoder1: I did tweak this to be terser, but I did want to differentiate redundantly between `allow -> request, denied -> response`

RichiCoder1:
Address most all the comments, added a touch more clarity, and went ahead and added an Output reference to match the Input reference.

mchmarny (approved):
/lgtm

PR #797, opened 2020-09-17
Add disclaimer about plaintext secrets

tcnghia:
## Description

Add a disclaimer similar to that of https://github.com/dapr/quickstarts/tree/master/hello-kubernetes#step-2---create-and-configure-a-state-store so that users know that plaintext secrets aren't the norm, and how they can do things more securely.


mchmarny (commented):
nit on verbiage 

> The above example uses secret in plain text, follow these instructions to configure secrets securely in production.

tcnghia:
> The above example uses secret in plain text

Done! Thanks for the feedback.

mchmarny (approved):
/lgtm

PR #810, opened 2020-09-24
Clarification on use of CloudEvents during publish

mchmarny:
Clarification on the use of CloudEvents during publish that resulted from this issue https://github.com/dapr/dapr/issues/2080


yaron2 (approved):
lgtm

Issue #832, opened 2020-09-28
AKS provisioning command using deprecated options/version


halspang:
## Expected Behavior

Command in [setup-aks](https://github.com/dapr/docs/blob/master/getting-started/cluster/setup-aks.md) to create the aks cluster should run without errors/warnings.

```
az aks create --resource-group [your_resource_group] --name [your_aks_cluster_name] --node-count 2 --kubernetes-version 1.14.7 --enable-addons http_application_routing --enable-rbac --generate-ssh-keys
```

## Actual Behavior

Running the command as it is currently in the docs yields an error that the kubernetes version is no longer supported and a warning that `--enable-rbac` is deprecated.

```
Operation failed with status: 'Bad Request'. Details: Version 1.14.7 is not supported in this region. Please use [az aks get-versions] command to get the supported version list in this region. For more information, please check https://aka.ms/supported-version-list
```

```
Argument 'enable_rbac' has been deprecated and will be removed in a future release. Use '--disable-rbac' instead.
```

## Steps to Reproduce the Problem

Run the command as given.

## Fix
Update the command to no longer use the `--kubernetes-version` option as the default value is above the required 1.13.x and remove the `--enable-rbac` option as this is the default for an aks cluster.

### Default versions per region
```
australiaeast: 1.17.11
australiasoutheast: 1.17.11
brazilsouth: 1.17.11
canadacentral: 1.17.11
canadaeast: 1.17.11
centralindia: 1.17.11
centralus: 1.17.11
eastasia: 1.17.11
eastus: 1.17.11
eastus2: 1.17.11
francecentral: 1.17.11
germanynorth: 1.17.11
germanywestcentral: 1.17.11
japaneast: 1.17.11
japanwest: 1.17.11
koreacentral: 1.17.11
koreasouth: 1.17.11
northcentralus: 1.17.11
northeurope: 1.17.11
norwayeast: 1.17.11
norwaywest: 1.17.11
southafricanorth: 1.17.11
southcentralus: 1.17.11
southeastasia: 1.17.11
southindia: 1.17.11
switzerlandnorth: 1.17.11
switzerlandwest: 1.17.11
uaecentral: 1.17.11
uaenorth: 1.17.11
uksouth: 1.17.11
ukwest: 1.17.11
westcentralus: 1.17.11
westeurope: 1.17.11
westus: 1.17.11
westus2: 1.17.11
```

PR #838, opened 2020-10-01
Fix aks cluster provisioning command

halspang:
## Description

When originally written, the command to provision an aks cluster
had to specify a specific version of Kubernetes. The requirement
of having Kubernetes version 1.13.x or higher is now met by the
default version of Kubernetes in all regions.

Additionally, all clusters now have rbac enabled by default so this
option is also removed.

## Issue reference

https://github.com/dapr/docs/issues/832


youngbupark (approved):
👍

PR #839, opened 2020-10-01
added FAQ on perf benchmark

msfussell:
Added FAQ on perf benchmark, since this question has been raised a few time in the community

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -1,6 +1,7 @@
 # FAQ
 
 - **[Networking and service meshes](#networking-and-service-meshes)**
+- **[Performance Benchmarks](#performance-benchmarks)**
 - **[Actors](#actors)**
 - **[Developer language SDKs and frameworks](#developer-language-sdks-and-frameworks)**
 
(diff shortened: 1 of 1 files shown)


FAQ.md
```diff
@@ -26,9 +27,12 @@ Read [How does Dapr work with service meshes?](https://github.com/dapr/dapr/wiki
 
 Istio is not a programming model and does not focus on application level features such as state management, pub-sub, bindings etc. That is where Dapr comes in.
 
+## Performance Benchmarks
+The Dapr project is focused on performance due to the inherent discussion of Dapr being a sidecar to your application. This [performance benchmark video](https://youtu.be/4kV3SHs1j2k?t=783) discusses and demos the work that has been done so far. The performance benchmark data is planned to be published on a regular basis along with the tests that you can run in your own test environment.
```
yaron2, line 31: The tests are already published for several months now. It's the data for the tests we (maintainers) ran on a specific configuration that needs to be published.

msfussell: Ok. Looks good. Can you approved this based on your changes?

yaron2: I didn't make the changes.

msfussell: Ok. So it says exactly that "The performance benchmark data is planned to be published on a regular basis along with the tests that you can run in your own test environment." 
No changes needed

yaron2: Ok, so my thinking was that it might taken that the tests themselves should be published along with the test data, which is not the case.

I suggest to rephrase to say that only the test benchmarks will be published.

msfussell: How about this change
"The performance benchmark data is planned to be published on a regular basis. You can also run the perf tests in your own environment to get perf numbers."

yaron2 (approved):
LGTM

PR #870, opened 2020-10-19
Add documentation on how to setup OpenTelemetry Collector

tcnghia:
## Description

Add documentation on how to setup OpenTelemetry Collector, using Application Insights as an example use case.

## Issue reference

Please reference the issue this PR will close: https://github.com/dapr/dapr/issues/1080


tcnghia:
@halspang @youngbupark can you please help take a look? Thanks!

concepts/observability/README.md
```diff
@@ -36,4 +36,5 @@ The observability tools listed below are ones that have been tested to work with
 ### Distributed Tracing
 
 * [How-To: Set up Zipkin](../../howto/diagnose-with-tracing/zipkin.md)
-* [How-To: Set up Application Insights](../../howto/diagnose-with-tracing/azure-monitor.md)
\ No newline at end of file
+* [How-To: Set up Application Insights with Local Forwarder](../../howto/diagnose-with-tracing/local-forwarder.md)
```
youngbupark, line 39: @msfussell do we want to remove local forwarder doc here? otherwise, I would add `[Deprecated]` comment on this link.

howto/diagnose-with-tracing/open-telemetry-collector.md
```diff
@@ -0,0 +1,254 @@
...
+
+> **NOTE**: You can register multiple exporters at the same time, and
+> the tracing logs are forwarded to all registered exporters.
+
+Deploy and run some applications. After a few minutes, you should see
+tracing logs appearing in your Application Insights resource. You can
+also use **Application Map** to examine the topology of your services,
+as shown below:
+
+![Application map](../../images/azure-monitor.png)
```
youngbupark, line 225: Can you see the same image when you use OTEL ? I would update this image as well..

tcnghia: Done

youngbupark:
image

OTEL spec has been changed - https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/database.md#connection-level-attributes

We need to update attributes to be compliant with this spec.

Annotation attributes keys: https://github.com/dapr/dapr/blob/0673a4af9a41cbc2cfb020fadacf3b60bc3d7329/pkg/diagnostics/tracing.go#L31-L61

OTEL trace attribute population logic for grpc : https://github.com/dapr/dapr/blob/0673a4af9a41cbc2cfb020fadacf3b60bc3d7329/pkg/diagnostics/grpc_tracing.go#L156

for HTTP: https://github.com/dapr/dapr/blob/0673a4af9a41cbc2cfb020fadacf3b60bc3d7329/pkg/diagnostics/http_tracing.go#L212

tcnghia:
I filed https://github.com/dapr/dapr/issues/2288 to track correcting the attribute keys, which should be fixed independently of this PR and http://github.com/dapr/dapr/issues/1080

tcnghia:
> I filed [dapr/dapr#2288](https://github.com/dapr/dapr/issues/2288) to track correcting the attribute keys, which should be fixed independently of this PR and [http://github.com/dapr/dapr/issues/1080](https://github.com/dapr/dapr/issues/1080)

Actually, we shouldn't add a doc to replace LocalForwarder with OpenTelemetry collector if the latter is currently broken. @msfussell do you agree? In that case, I think we should block this until https://github.com/dapr/dapr/issues/2288 is fixed.

howto/diagnose-with-tracing/open-telemetry-collector.md
```diff
@@ -0,0 +1,254 @@
...
+            path: /
+            port: 13133
+      volumes:
+        - configMap:
+            name: otel-collector-conf
+            items:
+              - key: otel-collector-config
+                path: otel-collector-config.yaml
+          name: otel-collector-config-vol
+EOF
```
msfussell, line 151: For consistency can you do these as apply this command to this file file rather than as single embedded "command and file. " 
For example  kubectl apply -f  opentelemetry.yaml

howto/diagnose-with-tracing/open-telemetry-collector.md
```diff
@@ -0,0 +1,254 @@
...
+  namespace: default
+spec:
+  # Configure Dapr tracing.
+  tracing:
+    samplingRate: "1"
+---
+apiVersion: dapr.io/v1alpha1
+kind: Component
+metadata:
+  name: native
```
msfussell, line 172: Is the component type really "native". This seems and odd name.
@youngbupark - Where did this name come from and should there be a better name?

howto/diagnose-with-tracing/open-telemetry-collector.md
```diff
@@ -0,0 +1,254 @@
...
+spec:
+  ...
+  template:
+    metadata:
+      ...
+      annotations:
+        dapr.io/enabled: "true"
+        dapr.io/app-id: "calculator-front-end"
+        dapr.io/app-port: "8080"
+        dapr.io/config: "tracing"
```
msfussell, line 205: This should be "appconfig" not "tracing" to match the example about, correct? 
This is a how to, so the steps have to all work together.

howto/diagnose-with-tracing/open-telemetry-collector.md
```diff
@@ -0,0 +1,254 @@
+# Using OpenTelemetry Collector to collect traces
+
+Dapr can integrate with [OpenTelemetry
+Collector](https://github.com/open-telemetry/opentelemetry-collector)
+through OpenCencus API. This guide will walkthrough an example to use
```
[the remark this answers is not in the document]
tcnghia: Done

howto/diagnose-with-tracing/open-telemetry-collector.md
tcnghia: I can do that to be consistent. However, that would require the users to be in a certain directory (checking out the docs in a repo) or will need to use a hyperlink. Using a hyperlink is probably more convenient.

@msfussell @orizohar @AaronCrawfis let me know which approach above is the preferred choice?

howto/diagnose-with-tracing/open-telemetry-collector.md
tcnghia: Yes, the component is really named `native`... it is a misnomer I believe.

howto/diagnose-with-tracing/open-telemetry-collector.md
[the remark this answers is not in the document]
tcnghia: @msfussell fixed this. Thanks for spotting.

tcnghia: Thanks for fixing this already.

howto/diagnose-with-tracing/open-telemetry-collector.md
AaronCrawfis: I recommend "create a file names `opentelemetry.yaml` with this content:

```yaml
CONTENTS
```

Apply this configuration with `kubectl apply -f opentelemetry.yaml`

tcnghia: @AaronCrawfis so, something like this?

1. Create a file named opentelemetry.yaml
```
cat > opentelemetery.yaml <<EOF
CONTENTS
EOF
```

2. Apply ...

Since the user will need to copy the content anyway, might as well have them copy a few more lines for convenience.

tcnghia: Updated. Please take a look.

tcnghia:
@halspang @AaronCrawfis @msfussell  I've converted the inlined YAMLs into files since it seems @halspang has trouble replicating the result. Having checked in files may prevent strange indentation issues.

AaronCrawfis (approved):

PR #878, opened 2020-10-26
Fix dead link to Ignite video session

discostu105:
The link to Mark Russinovich ignite session is dead. I've found the session on Youtube and replaced the link.


yaron2 (approved):
lgtm

PR #901, opened 2020-11-10
Amolenk/pubsubconsumerids

amolenk:
## Description

For the pub/sub building block components, the consumerID is always set to the application id by the runtime. Therefore, the consumerID property should be removed from configuration samples that still include it.

## Issue reference

This PR will close https://github.com/dapr/components-contrib/issues/521


yaron2 (approved):
lgtm

PR #906, opened 2020-11-11
Add versioning field for components

yaron2:
This PR adds the `version` field to all instances showing a Dapr `Component` manifest.


msfussell (approved):
LGTM

PR #915, opened 2020-11-13
Middleware pipelines: Fix link in next step section

tomzx:
## Description

This is a minor fix to the broken link in the middleware pipelines page.

Currently shows up as
[How-To: Configure API authorization with OAuth(https://docs.dapr.io/operations/security/oauth/)
instead of
[How-To: Configure API authorization with OAuth](https://docs.dapr.io/operations/security/oauth/)


AaronCrawfis (approved):
Thanks for the fix!

yaron2 (approved):
LGTM, thanks!

Issue #693, opened 2020-07-15
Distributed Tracing with Jaeger


cecilphillip:
## Describe the proposal
Currently, there's a reference for setting up zipkin for tracing. It would be nice to have a document showing how to use jaeger as well via the OpenTelemetry support.

msfussell:
This is a good idea. Now that OT is being integrated in the v1.0 RC releases, showing jaeger as an option makes sense

tcnghia:
@wcs1only @artursouza can you please add `RELEASE_NOTE: *UPDATED* docs to add Jaeger instructions` ? I can't edit this issue. Thanks

duncanworthy:
The original proposal was to add examples to show the use of Jaeger via OpenTelemetry support, however the completed PRs only showed how to configure Jaeger as a slot-in replacement for Zipkin using its support for the Zipkin API.
Is there any intent to address the original proposal of providing an example of setting up Jaeger via OpenTelemetry?

PR #950, opened 2020-11-24
Add Jaeger setup instruction

tcnghia:

## Description

Add instruction for using Dapr with Jaeger.

## Issue reference

_Please reference the issue this PR will close: https://github.com/dapr/docs/issues/693_


tcnghia:
@orizohar can you please help review?  thanks

daprdocs/content/en/operations/monitoring/jaeger.md
```diff
@@ -0,0 +1,106 @@
+---
+type: docs
+title: "How-To: Set up Jaeger for distributed tracing"
+linkTitle: "Jaeger"
+weight: 3000
+description: "Set up Jaeger for distributed tracing"
+type: docs
+---
+
+The following steps shows you how to configure Dapr to send distributed tracing data to Jaeger running as a container in your Kubernetes cluster, how to view them.
```
orizohar, line 10: Is this only for Kubernetes or can me make it more general (for example if someone wants to run a Jaeger container locally) if it has to be Kubernetes we should add it in the title

daprdocs/content/en/operations/monitoring/jaeger.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Setup
+
+First the following YAML file to install Jaeger
```
orizohar, line 14: Missing a verb? Maybe "First **create** the following YAML..."

daprdocs/content/en/operations/monitoring/jaeger.md
```diff
@@ -0,0 +1,106 @@
...
+  ingress:
+    enabled: false
+  allInOne:
+    image: jaegertracing/all-in-one:1.13
+    options:
+      query:
+        base-path: /jaeger
+```
+
+Now, let's use that Jaeger Operator resource to install Jaeger
```
orizohar, line 32: please remove use of first person. For example change to "Now, use the above YAML file to install Jaeger"

daprdocs/content/en/operations/monitoring/jaeger.md
```diff
@@ -0,0 +1,106 @@
...
+
+* jaeger.yaml component
+
+```yaml
+apiVersion: dapr.io/v1alpha1
+kind: Component
+metadata:
+  name: zipkin
+spec:
+  # Note that since we are using the Zipkin protocol to talk to Jaeger, the component
```
orizohar, line 53: I'd pull this comment out of the YAML file and explicitly explain this before this part, this will surely be a confusing part of the instruction. Also, elaborate if possible (Dapr currently supports two kinds of exporters, since Jaeger is compatible with Zipkin... and so on...)

daprdocs/content/en/operations/monitoring/jaeger.md
```diff
@@ -0,0 +1,106 @@
...
+```bash
+kubectl port-forward svc/jaeger-query 16686
+```
+
+In your browser, go to `http://localhost:16686` and you will see the Jaeger UI.
+
+![jaeger](/images/jaeger_ui.png)
+
+## References
+- [W3C distributed tracing]({{< ref w3c-tracing >}})
```
orizohar, line 106: Any external links we can add for using and configuring Jaeger?

orizohar (changes requested):

daprdocs/content/en/operations/monitoring/jaeger.md
tcnghia: Done

daprdocs/content/en/operations/monitoring/jaeger.md
tcnghia: Added

tcnghia (commented):
Addressed the feedback. Thanks @orizohar! Please take another look?

daprdocs/content/en/operations/monitoring/jaeger.md
```diff
@@ -0,0 +1,173 @@
+---
+type: docs
+title: "How-To: Set up Jaeger for distributed tracing"
+linkTitle: "Jaeger"
+weight: 3000
+description: "Set up Jaeger for distributed tracing"
+type: docs
+---
+Dapr currently supports two kind of tracing protocol: OpenCensus and
+Zipkin. Since Jaeger is compatible with Zipkin, we can use the Zipkin
```
orizohar, line 10: Please remove first person "we"

daprdocs/content/en/operations/monitoring/jaeger.md
```diff
@@ -0,0 +1,173 @@
...
+  ingress:
+    enabled: false
+  allInOne:
+    image: jaegertracing/all-in-one:1.13
+    options:
+      query:
+        base-path: /jaeger
+```
+
+Now, use the above YAML file to install install Jaeger
```
orizohar, line 98: double "install"

daprdocs/content/en/operations/monitoring/jaeger.md
```diff
@@ -0,0 +1,173 @@
...
+Zipkin. Since Jaeger is compatible with Zipkin, we can use the Zipkin
+protocol to talk to Jaeger.
+
+## Configure self hosted mode
+
+### Setup
+
+For self hosted mode, on running `dapr init` the following YAML files are created by default and they are referenced by default on `dapr run` calls unless otherwise overridden.
+
+1. The following file in `$HOME/.dapr/components/zipkin.yaml` or `%USERPROFILE%\dapr\components\zipkin.yaml`:
```
orizohar, line 19: I suggest avoiding the requirement to change the default components files that are created on install. Those should probably stay as is. I think we should direct users to create YAML files and then use them in the run command with `--components-path'. That is much cleaner and allows a user to have no side effects on the default Dapr dev env

daprdocs/content/en/operations/monitoring/jaeger.md
```diff
@@ -0,0 +1,173 @@
...
+spec:
+  type: exporters.zipkin
+  version: v1
+  metadata:
+  - name: enabled
+    value: "true"
+  - name: exporterAddress
+    value: "http://localhost:9412/api/v2/spans"
+```
+2. The following file in `$HOME/dapr/config.yaml` or `%USERPROFILE%\dapr\config.yaml`:
```
orizohar, line 41: What is the action the reader needs to do at this stage? is it to create this file and copy the contents below? please clarify
Also, it would better to clarify that the "or" refers to the OS you are running. Maybe something like this: 
`$HOME/dapr/config.yaml` (on Linux/MacOS) or `%USERPROFILE%\dapr\config.yaml` (on Windows)

daprdocs/content/en/operations/monitoring/jaeger.md
```diff
@@ -0,0 +1,173 @@
...
+
+```bash
+docker run -d --name jaeger \
+  -e COLLECTOR_ZIPKIN_HTTP_PORT=9412 \
+  -p 16686:16686 \
+  -p 9412:9412 \
+  jaegertracing/all-in-one:1.21
+```
+
+4. The applications launched with `dapr run` will by default reference the config file in `$HOME/dapr/config.yaml` or `%USERPROFILE%\dapr\config.yaml` and can be overridden with the Dapr CLI using the `--config` param:
```
orizohar, line 66: Same comment on clarifying OS

orizohar (changes requested):
Thanks for addressing my comments. Please see my followup comments

daprdocs/content/en/operations/monitoring/jaeger.md
tcnghia: Done

daprdocs/content/en/operations/monitoring/jaeger.md
tcnghia: That's a good idea. Fixed.

daprdocs/content/en/operations/monitoring/jaeger.md
tcnghia: I copied from zipkin.md, which may have more explanations to tell the users where the component files are.

Updated now to just create a new component file and use `--components-path` instead.

daprdocs/content/en/operations/monitoring/jaeger.md
tcnghia: Updated now to use `--components-path` which will not need variations like this.

daprdocs/content/en/operations/monitoring/jaeger.md
tcnghia: Fixed. Thanks for catching.

tcnghia (commented):
Thanks Ori. I've addressed the feedback.

Yes, using `--components-path` is cleaner and avoiding the OS-dependent paths.

orizohar (approved):
Thanks for this great addition to the docs and thank you for the quick response for feedback!

PR #958, opened 2020-11-28
Add 201 response for storage create

withinboredom:
## Description

While creating a new client, I discovered that 201 is also a potential response. Updated it here.


AaronCrawfis:
@wcs1only could you take a look and ensure this status code lines up with what's expected?

wcs1only:
We actually removed all 201 status codes returned in the following PR:
https://github.com/dapr/dapr/pull/2410

As of 1.0.0-rc.1, we should not be returning 201 anywhere. The 201 status code is used for resources created where a URL is returned in either a header or the body of the reply. We weren't doing this, so we adjusted our code accordingly.

Which Dapr runtime where you using? Version 1.0.0-rc.1 and onward should not be returning 201 anywhere.

Issue #891, opened 2020-11-03
Docs disappear when resizing brower windows


tcnghia:
## Expected Behavior

When resizing windows, doc texts get squeezed but still visible.

## Actual Behavior

Text disappears when the browser window approaches 8-inch width. 

## Steps to Reproduce the Problem

* Opening docs.dapr.io in Microsoft Edge browser
* Gradually reduce window width.

![ezgif-1-95852688d830](https://user-images.githubusercontent.com/138266/98044697-21158080-1ddc-11eb-8569-740b023faea8.gif)

## Release Note

RELEASE NOTE:

tcnghia:
cc @AaronCrawfis @orizohar

AaronCrawfis:
CSS issue is causing bootstrap columns to not resize properly. Content (middle column) is getting pushed below left column. Will look at when I get back from vacation on the 12th.

tcnghia:
Fixed by https://github.com/dapr/docs/pull/980

PR #980, opened 2020-12-03
Removed min-width to fix resizing issue

AaronCrawfis:
Addresses #891 

Removed min-width and used padding instead


orizohar (approved):
should this also go into master and rc.1?

AaronCrawfis:
> should this also go into master and rc.1?

Yep, about to do that now

Issue #978, opened 2020-12-03
Update trace spec in Configuration to mention `zipkin` section


tcnghia:
**What content needs to be created or modified?**
As part of https://github.com/dapr/dapr/issues/2337 ps://we need to update the doc here https://docs.dapr.io/operations/configuration/configuration-overview/#tracing to mention the new `zipkin` section (added for https://github.com/dapr/dapr/issues/2521)

**Describe the solution you'd like**

Property | Type | Definition
-- | -- | --
zipkin.endpointAddress | string | Set end point address for the Zipkin ingestion.

**Where should the new material be placed?**
https://docs.dapr.io/operations/configuration/configuration-overview/#tracing

**Additional context**
 https://github.com/dapr/dapr/issues/2337

PR #983, opened 2020-12-04
Document the new fields in Configuration's tracing section.

tcnghia:
## Description

Document the new fields in Configuration's tracing section. This is a follow up of https://github.com/dapr/dapr/pull/2532 .

## Issue reference

Fixes https://github.com/dapr/docs/issues/978

### DIFF
--- a/daprdocs/content/en/operations/configuration/configuration-overview.md
+++ b/daprdocs/content/en/operations/configuration/configuration-overview.md
@@ -15,7 +15,7 @@ In self hosted mode the Dapr configuration is a configuration file, for example 
 
 A Dapr sidecar can also apply a configuration by using a ```--config``` flag to the file path with ```dapr run``` CLI command.
 
-#### Kubernetes sidecar 
+#### Kubernetes sidecar
 In Kubernetes mode the Dapr configuration is a Configuration CRD, that is applied to the cluster. For example;
 
 ```bash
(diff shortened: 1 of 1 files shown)


tcnghia:
@orizohar @msfussell can you please help take a look at this small PR? thanks!

msfussell:
One thing that it would be good to add as part of this PR is to be clear that metrics can also be included as its own section. It is explicitly being missed here. It is in this doc https://v1-rc1.docs.dapr.io/developing-applications/building-blocks/observability/metrics/ but it would be good to have this all in one place.

So can you add metrics?
Otherwise this looks good.

Like this
apiVersion: dapr.io/v1alpha1
kind: Configuration
metadata:
  name: tracing
  namespace: default
spec:
  tracing:
    samplingRate: "1"
  metric:
    enabled: false

tcnghia:
> One thing that it would be good to add as part of this PR is to be clear that metrics can also be included as its own section. It is explicitly being missed here. It is in this doc https://v1-rc1.docs.dapr.io/developing-applications/building-blocks/observability/metrics/ but it would be good to have this all in one place.
> 
> So can you add metrics?
> Otherwise this looks good.
> 
> Like this
> apiVersion: dapr.io/v1alpha1
> kind: Configuration
> metadata:
> name: tracing
> namespace: default
> spec:
> tracing:
> samplingRate: "1"
> metric:
> enabled: false

Done now

msfussell (approved):
LGTM

Issue #967, opened 2020-12-01
Update trace configurations to not use exporter component


tcnghia:
**What content needs to be created or modified?**
As proposed in https://github.com/dapr/dapr/issues/2337, we will be moving trace configurations into Dapr core and deprecate the trace exporter component. 

As a result we need to update related documentation.

**Describe the solution you'd like**
- [ ] When trace setting is available in **Configuration**, all documentation should use that. Also, since before OpenTelemetry GA, Zipkin is the only protocol (others can be supported by using OpenTelemetry Collector to translate Zipkin).
- [ ] After new docs are available we want to announce at the release announcement

**Additional context**
See whole plan at https://github.com/dapr/dapr/issues/2337

PR #990, opened 2020-12-08
Update OpenTelemetry Collector doc to use Zipkin API

tcnghia:
## Description

* Update OpenTelemetry Collector doc to use Zipkin API: As part of https://github.com/dapr/dapr/issues/2337, we are updating the OpenTelemetry Collector doc to use Zipkin API instead of OpenCensus API.

## Issue reference

Part of https://github.com/dapr/docs/issues/967


daprdocs/static/docs/open-telemetry-collector/collector-config.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: dapr.io/v1alpha1
+kind: Configuration
+metadata:
+  name: appconfig
+  namespace: default
+spec:
+  tracing:
+    samplingRate: "1"
+    zipkin:
+      endpointAddress: "http://otel-collector.default.svc.cluster.local:9411/api/v2/spans"
```
artursouza, line 10: This might be too late for this but why not simply `endpoint`?

tcnghia: That's a good call. It's not too late.

cc @msfussell I like this suggestion here ^^^. We can still rename things now. WDYT

msfussell: I prefer to keep endpointAddress, since this is descriptive and there is no real need to shorten this.

tcnghia:
@orizohar @youngbupark can you please help review this PR? thanks!

youngbupark (approved):

tcnghia:
@youngbupark @orizohar can you please help merge the PR? thanks

PR #992, opened 2020-12-08
Update CLI reference

### LINKED ISSUE #972: Update CLI reference
Update CLI reference based on changes in https://github.com/dapr/cli/pull/541

mukundansundar:

## Description

Update CLI reference for Milestone 2. 

## Issue reference

_Please reference the issue this PR will close: https://github.com/dapr/docs/issues/972


daprdocs/content/en/reference/cli/dapr-mtls--export.md
```diff
@@ -0,0 +1,29 @@
+---
+type: docs
+title: "mtls export CLI command reference"
+linkTitle: "mtls export"
+description: "Detailed information on the mtls export CLI command"
+---
+
+## Description
+
+Export the root CA, issuer cert and key from Kubernetes to local files
```
msfussell, line 10: Should this mtls export say "only from K8s"? Should this not also be supported platforms? Others could be supported one day and this could be applied locally if the sentry service was running locally as described here See https://docs.dapr.io/operations/security/mtls/#self-hosted

So maybe the text should be like this?
"Export the root CA, issuer cert and key from Dapr Sentry service to local files"

and the reason here is that the Sentry service does run locally.

daprdocs/content/en/reference/cli/dapr-mtls-expiry.md
```diff
@@ -0,0 +1,28 @@
+---
+type: docs
+title: "mtls expiry CLI command reference"
+linkTitle: "mtls expiry"
+description: "Detailed information on the mtls expiry CLI command"
```
msfussell, line 5: This is correct in that this is not K8s specific. See previous comment on mtls export.

daprdocs/content/en/reference/cli/dapr-run.md
```diff
@@ -17,19 +17,26 @@ dapr run [flags] [command]
 
 ## Examples
```
msfussell, line 19: Can we add a dotnet example here, to get dotnet some love! Like this

dapr run --app-id myapp --app-port 5000 -- dotnet run

daprdocs/content/en/reference/cli/dapr-mtls--export.md
mukundansundar: right now the command does only run for Kubernetes https://github.com/dapr/cli/blob/master/pkg/kubernetes/mtls.go. 

I think we can change the help message when we start supporting other platforms.

daprdocs/content/en/reference/cli/dapr-mtls-expiry.md
mukundansundar: no this is kubernetes specific. see the linked go file in github above.

daprdocs/content/en/reference/cli/dapr-init.md
```diff
@@ -7,14 +7,30 @@ description: "Detailed information on the init CLI command"
...
+```
+
+### Initialize Dapr in Kubernetes
+```bash
+dapr init -k
+```
+
+### Initialize Dapr in slim self-hosted mode
+```bash
+dapr init -s
```
msfussell, line 31: can you provide an example here of installing a particular version of the dapr runtime? use v1.0.0 ;-)

dapr init --runtime-version 1.0.0

msfussell (commented):
Added comments to address

msfussell (dismissed):
LGTM

msfussell (dismissed):
Approve

AaronCrawfis (dismissed):

AaronCrawfis (approved):

Issue #851, opened 2020-10-08
Update the mTLS configuration doc to describe how to run Sentry control plan service locally


msfussell:
This doc https://github.com/dapr/docs/blob/3ac61f2933edf858dc97219dfb7b82d619fd9831/howto/configure-mtls/README.md#self-hosted need updates to describe how to start the Sentry service locally so that developers can test the allowlist feature.

The https://github.com/dapr/docs/tree/master/howto/allowlists-serviceinvocation doc needs to be updated at the start of the doc to point to the above doc . In other words we need to make it clear that allowlists depends upon the Sentry service to work correctly.

AaronCrawfis:
Related to #857

yaron2:
Closed with #996

PR #996, opened 2020-12-09
update steps to run hello-world with ACL in selfhosted mode

vinayada1:

## Description

Updated the steps to run hello world sample with Access control list in self hosted mode.

#851 

_Please reference the issue this PR will close: #851


yaron2 (commented):
Looks good overall, I'm wondering if we should remove the pythonapp / nodeapp specific bits, as this is not a quickstart...

vinayada1:
> Looks good overall, I'm wondering if we should remove the pythonapp / nodeapp specific bits, as this is not a quickstart...

However, this is in context of running the hello-world sample. I updated some links to point to this.

msfussell:
Did quite a bit of re-writing and restructuring of the docs. Please read this through.  It is important that our instructions work for both Linux/MacOs and Windows and so you need to provide env configuration settings for Windows also. Hopefully the cert file exported by Sentry can be reference with these env variables will work for Windows also. Can you test this out?

You will see in the doc, tab headings for Linux/MacOS/Windows as in the example here https://docs.dapr.io/getting-started/install-dapr-cli/ 

I also structured this so there were example config files at the start and these were then used both locally and K8s, rather than having these all bundles under the K8s examples.

Note for future doc guidelines. 
1) References to other docs are now in the Hugo format like this [Init Dapr locally]({{< ref install-dapr.md >}})
2) Stay away from "we" usage, since this is unclear who/what this is referring to. Stick to you, wherever possible. https://docs.dapr.io/contributing/contributing-docs/#style-and-tone

daprdocs/content/en/operations/configuration/invoke-allowlist.md
```diff
@@ -233,7 +234,78 @@ spec:
...
+```bash
+export DAPR_TRUST_ANCHORS=`cat $HOME/.dapr/certs/ca.crt`
+export DAPR_CERT_CHAIN=`cat $HOME/.dapr/certs/issuer.crt`
+export DAPR_CERT_KEY=`cat $HOME/.dapr/certs/issuer.key`
+export NAMESPACE=default
+ ```
+{{% /codetab %}}
+
+{{% codetab %}}
+```powershell
```
AaronCrawfis, line 285: Why are the Windows variables empty? Do they just need to be filled in?

daprdocs/content/en/operations/configuration/invoke-allowlist.md
```diff
@@ -233,7 +234,78 @@ spec:
...
+set DAPR_TRUST_ANCHORS=``
+set DAPR_CERT_CHAIN=``
+set DAPR_CERT_KEY=``
+set NAMESPACE=default
+```
+{{% /codetab %}} 
+6. Run daprd to launch a Dapr sidecar for the python app with mTLS enabled, referencing the local Sentry service.
```
AaronCrawfis, line 292: Why do we launch the sidecar separately from the app instead of using dapr run?

msfussell: This is an unfortunate limitation of dapr cli not having access to the Sentry service, which can only be set at the daprd level. It makes testing very aukward and I meant to raise an issue on this in the CLI for future consideration. It just depends on how common using the sentry service locally becomes.

yaron2: Locally we can expect it to never be used at all, there's simply no reason. It's expected to be used in a distributed environment, which doesn't preclude us adding the Sentry address flag to the CLI.

AaronCrawfis: Makes sense. We should raise that issue and place a note here in the docs that the sidecar needs to be launched separately for this reason.

vinayada1: Created issue: https://github.com/dapr/cli/issues/550

daprdocs/content/en/operations/configuration/invoke-allowlist.md
vinayada1: I added the powershell commands

msfussell (approved):
LGTM

PR #1012, opened 2020-12-15
Clarifies consumer groups and multiple instances

mchmarny:
Clarifies the behaviour around multiple app instances in a consumer groups


yaron2 (approved):
lgtm

PR #1069, opened 2021-01-03
updated rabbitmq pub/sub documentation

trondhindenes:

## Description
Updated rabbitmq documentation to include changes made in https://github.com/dapr/components-contrib/pull/562. Also changes the format to use the new https://v1-rc2.docs.dapr.io/operations/components/component-schema/ one.

## Issue reference
N/A

### DIFF
--- a/daprdocs/content/en/operations/components/setup-pubsub/supported-pubsub/setup-rabbitmq.md
+++ b/daprdocs/content/en/operations/components/setup-pubsub/supported-pubsub/setup-rabbitmq.md
@@ -5,7 +5,53 @@ linkTitle: "RabbitMQ"
 description: "Detailed documentation on the RabbitMQ pubsub component"
 ---
 
-## Setup RabbitMQ
+## Format
+
+```yaml
+apiVersion: dapr.io/v1alpha1
+kind: Component
+metadata:
+  name: rabbitmq-pubsub
+  namespace: default
+spec:
+  type: pubsub.rabbitmq
+  version: v1
+  metadata:
+  - name: host
+    value: "amqp://localhost:5672"
+  - name: durable
+    value: "false"
+  - name: deletedWhenUnused
+    value: "false"
+  - name: autoAck
+    value: "false"
+  - name: deliveryMode
+    value: "0"
+  - name: requeueInFailure
+    value: "false"
+  - name: prefetchCount
+    value: "0"
+```
+{{% alert title="Warning" color="warning" %}}
+The above example uses secrets as plain strings. It is recommended to use a secret store for the secrets as described [here]({{< ref component-secrets.md >}}).
+{{% /alert %}}
+
+## Fields
+
+| Field              | Required | Details | Example |
+|--------------------|:--------:|---------|---------|
+| host               | Y        | Connection-string for the rabbitmq host  | `amqp://user:pass@localhost:5672`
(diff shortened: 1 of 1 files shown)


trondhindenes:
btw I'm not sure if I should include the "global" fields (`apiVersion, kind, metadata.name`) etc in this PR. They're in the component schema (https://v1-rc2.docs.dapr.io/operations/components/component-schema/), but it's not quite clear to me if every component documentation should copy this. It would be awesome to have "template" to look at for component documentation.

AaronCrawfis:
@trondhindenes did the change to components-contrib make it into the rc2 release (and therefore applies to the rc2 docs)? Or is it only applicable to the upcoming rc3 release and docs?

AaronCrawfis:
Oh I actually am just now reading your note on [Discord](https://discord.com/channels/778680217417809931/778684392893579267/795306871770185749) regarding this exact problem. Will get back shortly on that.

trondhindenes:
> did the change to components-contrib make it into the rc2 release?

I'm pretty sure it targets rc3, but I'm not sure how to find out for certain.

yaron2:
~It made it into rc.2~

yaron2:
Sorry, this was NOT included in rc.2.

yaron2:
@trondhindenes please resolve the conflict, which you will find familiar :)

msfussell:
This format is very nice. Let's start rolling this out across all components. We should include all the global fields as developers should be able to cut and past the file and just use this with the Metadata settings of their choice.  I especially like the examples column, which is often where people get stuck

Some comments on your changes.
1) deleteWhenUnused defaults to "true" not "false" as set the table.
2) The Durable property is not in the metadata struct, so should be removed from the table.
3) The default reconnectWait time is 10s. This is not mentioned as the default in the table.

trondhindenes:
@msfussell thanks!
I totally agree that the yaml codeblock should be "copyable" directly. My question was whether or not it makes sense to document the "global" fields (`apiVersion`, `kind` etc) in the fields table. My opinion is that no, that doesn't make sense as it only bloats the table (this is also where this PR differs from the https://v1-rc2.docs.dapr.io/operations/components/component-schema/).

Thanks for looking thru the actual content - will correct.

msfussell:
@trondhindenes - Yes, do not include the general component schema, simply reference this in the related links like this - [Basic schema for a Dapr component]({{< ref component-schema >}})

can you push the suggest changes so that we can merge this PR?

msfussell (commented):
@trondhindenes 
Add - [Basic schema for a Dapr component]({{< ref component-schema >}}) in the Related links section
Change "Fields" heading to "Spec metadata fields"

msfussell:
@trondhindenes  - Any chance you can push an update to this PR, since we would like to get this in. 
You can see your new proposed format here https://v1-rc3.docs.dapr.io/operations/components/setup-pubsub/supported-pubsub/setup-redis-pubsub/

msfussell:
@trondhindenes - I updated the PR in your branch for all the changes. You just need to review these and ACK. 
This PR goes with this issue https://github.com/dapr/docs/issues/1084

trondhindenes:
@msfussell sorry for being so slow on this one. I'm fine with the changes you've made. I've resolved the conflict, and also added documentation for the concurrency setting that @yaron2 added a few days ago. Hopefully this is ready to merge now!

AaronCrawfis (approved):

PR #1104, opened 2021-01-21
Update kubernetes-annotations.md

apuzyrevsky:
adding new annotation for added functionality in PR https://github.com/dapr/dapr/pull/2679

## Description
Adding dapr.io/http-max-request-size annotation to dapr solution to define custom fasthttp server MaxRequestBodySize parameter to handle uploading of big files.


orizohar (approved):
Thanks for updating the docs with the new annotation!

PR #1122, opened 2021-01-27
Rework Azure Blob Storage documentation to be more clear

XavierGeerinck:

## Description

Clarify the documentation for Azure Blob Storage


daprdocs/content/en/operations/components/setup-bindings/supported-bindings/blobstorage.md
```diff
@@ -40,20 +40,62 @@ The above example uses secrets as plain strings. It is recommended to use a secr
...
 
+> Note: by default, a random UUID is generated. See below for Metadata support to set the name
+
 ```json
 {
   "operation": "create",
-  "data": {
-    "field1": "value1"
-  }
+  "data": "YOUR_CONTENT"
```
orizohar, line 48: As long as we are clarifying, can you please add a comment right after this about the fact that "YOUR_CONTENT" can be different formats (that you show further down)?

orizohar (dismissed):
Thanks for adding more examples and clarifying! Please see my one comment

daprdocs/content/en/operations/components/setup-bindings/supported-bindings/blobstorage.md
Proximus-Azure-Services: very valid comment! Would love to do so but I am working as well to get it working myself :) the issue I am having here is that a REST endpoint requires a string to be passed (due to the nature of { "data": CONTENT } and it's serialization. 

How do you see this? When working with images for example, I am seeing that I should pass them as Base64 and decode them with an internal function (which I can't seem to get working but that's another discussion)

XavierGeerinck: very valid comment! Would love to do so but I am working as well to get it working myself :) the issue I am having here is that a REST endpoint requires a string to be passed (due to the nature of { "data": CONTENT } and it's serialization.

How do you see this? When working with images for example, I am seeing that I should pass them as Base64 and decode them with an internal function (which I can't seem to get working but that's another discussion)

yaron2 (approved):
lgtm

Issue #1128, opened 2021-01-28
Need examples of acceptable syntax and formats for .PFX file path name in Az KeyVault Secret Store component


georgestevens99:
**What content needs to be created or modified?**
The Azure Key Vault Secret Store component definition requires a name value pair for the certificate .pfx file: 
- name: spnCertificateFile
    value : "[pfx_certificate_file_local_path]"

Please provide examples of acceptable paths.  Like fully qualified path, or just the .pfx file name in the .components directory (which does not seem to work -- does it always have to be a fully qualified path?). 

Please provide examples of acceptable path syntax.  For Linix based systems I assume C:/something/somethingelse/ttt.pfx works, but this will not work on Windows systems which needs C:\\something\\somethingelse\\ttt.pfx.

Make the examples you furnish so complete that it is so easy write a key vault component definition that one can't possibly make a mistake, even a stupid one!

**Describe the solution you'd like**
See above.

**Where should the new material be placed?**
Place this just under the text of the component definition.

**Additional context**

msfussell:
Fixed

PR #1132, opened 2021-01-29
Need examples of acceptable syntax and formats for .PFX file path name in Az KeyVault Secret Store component

georgestevens99:

- [ ] [Read the contribution guide](https://docs.dapr.io/contributing/contributing-docs/)

## Description

  I have not made any changes since I have never used the documents technology.  But here is what needs to happen, copied from Issue #1128.

The Azure Key Vault Secret Store component definition requires a name value pair for the certificate .pfx file:

name: spnCertificateFile
value : "[pfx_certificate_file_local_path]"

For the above value field, please provide examples of acceptable paths. Like fully qualified path, or just the .pfx file name in the .components directory (which does not seem to work -- **does it always have to be a fully qualified path**?  If so, say that.).

Please provide examples of **acceptable path syntax**. For Linix based systems I assume C:/something/somethingelse/ttt.pfx works, but this will not work on Windows systems which needs C:\something\somethingelse\ttt.pfx with each backslash escaped with another backslash.  Any single backslash produces an error.

## Issue reference

  Issue #1128


orizohar:
@georgestevens99 - It looks like this PR bunches a lot of things from branch v0.11 to v1.0-rc3 and includes 23 files. If there is a specific change you'd like to propose to address the issue you mention please open a PR with that change only.

orizohar:
Closing as the PR includes unrelated change and should come from a dev branch. Please open a separate PR to address specific issue mentioned for the correct target branch

PR #1227, opened 2021-02-16
Adding instrunctions to enable Dynatrace support for observing Dapr a…

dtPaTh:
## Description

Adding instructions for monitoriing support with Dynatrace.  Following the menu structure we separated instructions for metrics and traces and created a child page within metrics as well as traces/common backends


daprdocs/content/en/operations/monitoring/tracing/supported-tracing-backends/dynatrace.md
```diff
@@ -0,0 +1,83 @@
...
+
+[Dynatrace](https://www.dynatrace.com) OneAgent is responsible for collecting all telemetry within your monitored environment. 
+
+It automatically instruments your applications built with Dapr without the need for further configuration. 
+
+![Distributed Trace](/images/dt-waterfall.png)
+
+![Codelevel-Trace](/images/dt-ppath.png)
+
+## Installation
```
orizohar, line 17: I'm not seeing any guidance regarding the required Dapr configuration as in the guidance for other backends (e.g. where to point the Dapr traces). Is this missing? if not needed due to the OneAgent I think it still needs to be explicitly mentioned as it may be confusing to the reader

orizohar:
@dtPaTh - Thanks for this! Really well written and easy to follow. Thank you for following the guidance in the contribution docs. I had one comment in the tracing file

@tcnghia - can you please review as well?

daprdocs/content/en/operations/monitoring/tracing/supported-tracing-backends/dynatrace.md
```diff
@@ -0,0 +1,83 @@
...
+type: docs
+title: "How-To: Set up Dynatrace OneAgent for distributed tracing"
+linkTitle: "Dynatrace"
+weight: 1000
+description: "Set up Dynatrace OneAgent for distributed tracing and deep code-level insights in Kubernetes"
+---
+
+[Dynatrace](https://www.dynatrace.com) OneAgent is responsible for collecting all telemetry within your monitored environment. 
+
+It automatically instruments your applications built with Dapr without the need for further configuration.
```
tcnghia, line 11: Can you please explain in further detail how this happens? Does Dynatrace need to intercept the requests sent/received by the applications?

Currently, Dapr does that and then send trace to a Zipkin-supported URL, but it seems your sample didn't need to do that.

orizohar:
@dtPaTh - we'd like to get this PR in, can you please address the review comments?

xendren:
It looks like this document change was never implemented.  It would be extremely helpful if you mention this on the Observability page of the docs that Dynatrace OneAgent handles ALL metrics including from the Dapr sidecar.  As a Dapr and Dynatrace user, this is very important information for me.

PR #1232, opened 2021-02-17
Adding a few updates for versioning

msfussell:
The middleware components did not have version number in the table.
Also added a link to the Component Contrib authoring guidance in the repo from the versioning doc


pkedy (approved):
LGTM

This might be more my personal preference but I'm thinking the version column should come before the status column.

Thought for the future: Much like how the Version 2 and beyond doc explains, components and middleware have a `.md` suffix for v0-v1 and a `-v2.md` for version 2, etc.

yaron2 (approved):

orizohar (approved):

PR #1302, opened 2021-03-10
Minor fixes for sdk-serialization.md

aquamme:
This fixes a couple of typos, and what I presume is an outdated example.


daprdocs/content/en/developing-applications/sdks/sdk-serialization.md
```diff
@@ -149,6 +149,6 @@ redis-cli MGET "ActorStateIT_StatefulActorService||StatefulActorTest||1581130928
  redis-cli MGET "ActorStateIT_StatefulActorService||StatefulActorTest||1581130928192||mydata
 "eyJ2YWx1ZSI6Ik15IGRhdGEgdmFsdWUuIn0="
 ```
-6. When serializing a object that is a `byte[]`, the serializer should just pass it through since `byte[]` shoould be already handled internally in the SDK. The same happens when deserializing to `byte[]`.
+6. When serializing an object that is a `byte[]`, the serializer should just pass it through since `byte[]` should be already handled internally in the SDK. The same happens when deserializing to `byte[]`.
```
artursouza: Delete this comment for now, please. This behavior is under discussion now.

artursouza (changes requested):

artursouza (approved):

PR #1329, opened 2021-03-22
Additional cloud storage binding examples

bhummerstone:

## Description

I noticed that the documentation for S3 and GCP Storage were lacking examples, so I've copied the format from Azure Blob to fill this gap.

That said, it is frustrating that this isn't standardised: the Azure binding uses "blobName", the AWS binding uses "key" and the GCP binding uses "name", all to refer to the same metadata; might be something worth addressing in the future.

## Issue reference

No related issue


orizohar:
@bhummerstone - thanks for the PR adding the examples. Regarding your comment on standardizing - that is a fair point. To make sure this is considered and tracked, can you please open an issue for this in the dapr/dapr repo?

Better yet, if you can include a proposal in the issue that would be best.

bhummerstone:
Thanks @orizohar: have created the issue here: https://github.com/dapr/dapr/issues/2966

orizohar (approved):

PR #1331, opened 2021-03-23
Added details on the application-level dapr.io/enable-metrics annotation

### LINKED ISSUE #1299: [CONTENT] Mention new dapr.io/enable-metrics annotation
Per dapr/dapr#2900, the new `dapr.io/enable-metrics` annotation should be mentioned on the [metrics page](https://docs.dapr.io/operations/monitoring/metrics/metrics-overview/).

pkedy:
## Description

Added details and example for using the application-level `dapr.io/enable-metrics` annotation.

## Issue reference

Resolves #1299

### DIFF
--- a/daprdocs/content/en/operations/monitoring/metrics/metrics-overview.md
+++ b/daprdocs/content/en/operations/monitoring/metrics/metrics-overview.md
@@ -3,7 +3,7 @@ type: docs
 title: "Metrics"
 linkTitle: "Metrics"
 weight: 4000
-description: "Observing Dapr metrics"
+description: "Observing Dapr metrics in Kubernetes"
 ---
 
 Dapr exposes a [Prometheus](https://prometheus.io/) metrics endpoint that you can scrape to gain a greater understanding of how Dapr is behaving and to setup alerts for specific conditions.
@@ -34,7 +34,8 @@ spec:
         dapr.io/enabled: "true"
         dapr.io/app-id: "nodeapp"
         dapr.io/app-port: "3000"
-        dapr.io/enable-metrics: "false"
+        dapr.io/enable-metrics: "true"
+        dapr.io/metrics-port: "9090"
     spec:
       containers:
       - name: node
--- a/daprdocs/content/en/operations/monitoring/metrics/metrics-overview.md
+++ b/daprdocs/content/en/operations/monitoring/metrics/metrics-overview.md
@@ -57,7 +57,7 @@ spec:
   tracing:
     samplingRate: "1"
   metric:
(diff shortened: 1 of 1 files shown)


daprdocs/content/en/operations/monitoring/metrics/metrics-overview.md
```diff
@@ -12,9 +12,39 @@ Dapr exposes a [Prometheus](https://prometheus.io/) metrics endpoint that you ca
...
+      app: node
+  template:
+    metadata:
+      labels:
+        app: node
+      annotations:
+        dapr.io/enabled: "true"
+        dapr.io/app-id: "nodeapp"
+        dapr.io/app-port: "3000"
+        dapr.io/enable-metrics: "false"
```
orizohar: Is there an annotation to set the metric port as well? if so, I think an example with the port annotation and enabled=true would be better.

daprdocs/content/en/operations/monitoring/metrics/metrics-overview.md
```diff
@@ -12,9 +12,39 @@ Dapr exposes a [Prometheus](https://prometheus.io/) metrics endpoint that you ca
 
 The metrics endpoint is enabled by default, you can disable it by passing the command line argument `--enable-metrics=false` to Dapr system processes.
 
-The default metrics port is `9090`. This can be overridden by passing the command line argument `--metrics-port` to Daprd.
+The default metrics port is `9090`. This can be overridden by passing the command line argument `--metrics-port` to Daprd. Additionally, the metrics exporter can be disabled for a specific application by setting the `dapr.io/enable-metrics: "false"` annotation to your application deployment. With the metrics exporter disabled, `daprd` will not open the metrics listening port.
 
-To disable the metrics in the Dapr side car, you can use the `metric` spec configuration and set `enabled: false` to disable the metrics in the Dapr runtime.
+```yaml
```
orizohar, line 17: This whole article has an implicit assumption that the hosting environment is Kubernetes. Even though this probably only makes sense today for k8s I think it is worth mentioning in the first sentence in the page. Possibly add "... in Kubernetes" in the description field

orizohar (changes requested):

daprdocs/content/en/operations/monitoring/metrics/metrics-overview.md
pkedy: The default is actually true and thus usually not specified in the examples to date...  but I see what you are saying. I'll make the change and we can always change it back later.

orizohar (approved):

PR #1374, opened 2021-04-04
Add Scoop package manager instructions for Dapr

JadynWong:
Adding scoop installation method for windows platform

## Description

Adding a way to install cli using scoop for windows platform.

## Issue reference

https://github.com/ScoopInstaller/Main/pull/1839

### DIFF
--- a/daprdocs/content/en/getting-started/install-dapr-cli.md
+++ b/daprdocs/content/en/getting-started/install-dapr-cli.md
@@ -23,6 +23,11 @@ This Command Prompt command installs the latest windows Dapr cli to `C:\dapr` an
 ```powershell
 powershell -Command "iwr -useb https://raw.githubusercontent.com/dapr/cli/master/install/install.ps1 | iex"
 ```
+
+Or you can install via [Scoop](https://scoop.sh):
+```powershell
+scoop install dapr-cli
+```
 {{% /codetab %}}
 
 {{% codetab %}}


AaronCrawfis:
Thank you for your contribution! I am all for adding additional package managers to make installing Dapr easy for every developer.

However, my only concern is maintainability and keeping Scoop up to date as we release future cli versions. As it's currently implemented at https://github.com/ScoopInstaller/Main/blob/5b50b565ba18942d42983f5e10c899633e1cf9a6/bucket/dapr-cli.json, this points to version 1.1. When we release 1.1.1, or 1.2, who will be in charge of updating the Scoop manifest? Is the expectation that we should add it to our endgame process?

@yaron2 @artursouza @orizohar thoughts on this?

artursouza (changes requested):
We cannot add this to the official documentation unless this is being updated automatically by Dapr's official pipelines. This is a matter of security. Users of Dapr must have confidence that what they are consuming is being generated from the official Dapr maintainers.

We cannot block people from adding installers in different package managers but we should not document anything not maintained by this project.

For example, we do it for Homebrew: https://github.com/dapr/homebrew-tap

Our Homebrew config is automatically updated on every CLI release and we troubleshoot it in case it fails. There is nothing stopping someone to have their own homebrew "tap" with another Dapr CLI config but it would not be documented here.

JadynWong:
@AaronCrawfis 
It is will be auto check update pre hour.

@artursouza 
As you said, scoop's buckets management is not the same as homebrew's formulae. Although it supports adding self-hosted buckets, I think it will increase your maintenance cost and the usage is not as user-friendly as homebrew.

Based on your instructions, I will close this PR.Thanks.

artursouza:
> @AaronCrawfis
> It is will be auto check update pre hour.
> 
> @artursouza
> As you said, scoop's buckets management is not the same as homebrew's formulae. Although it supports adding self-hosted buckets, I think it will increase your maintenance cost and the usage is not as user-friendly as homebrew.
> 
> Based on your instructions, I will close this PR.Thanks.


Increasing the maintenance cost is OK but it must be automated, so we just "maintain" if it breaks. I am open to have a Scoop bucket for Dapr but it must be done in an official way by integrating it into our release pipeline.

Thanks for the contribution.

Issue #1336, opened 2021-03-24
https://docs.dapr.io/getting-started/configure-redis/#create-a-redis-store points to docs in Chinese


vinayada1:
**Describe the bug**
https://docs.dapr.io/getting-started/configure-redis/#create-a-redis-store points to Chinese docs.

**Steps to reproduce**
1. Go to https://github.com/dapr/quickstarts/tree/master/pub-sub
2. Go to the paragraph below and click the "these steps" link
```
Follow these steps to create a Redis store using Helm.

Note: Currently the version of Redis supported by Azure Redis Cache is less than 5, so using Azure Redis Cache will not work.
```
3. It points to docs in Chinese

**Expected behavior**
The default language should be English

**Screenshots**

**Desktop (please complete the following information):**
 - OS: 
 - Browser 
 - Version 

**Smartphone (please complete the following information):**
 - Device: 
 - OS: 
 - Browser 
 - Version 

**Additional context**

orizohar:
@AaronCrawfis - My suspicion is that this has to do with Hugo aliases. The above link is an old link, if the Chinese page defines it as an alias without the /zh-hans/ path it will "hijack" the English page

orizohar:
Also happens with [https://docs.dapr.io/getting-started/install-dapr/#install-dapr-on-a-kubernetes-cluster](https://docs.dapr.io/getting-started/install-dapr/#install-dapr-on-a-kubernetes-cluster)

PR #1377, opened 2021-04-05
Update Chinese translated content to v1.1

newbe36524:

## Description

- update latest source content from  v1.1
- update latest translation to v1.1


newbe36524:
@skyao could you please help to review this content ? many thanks.

AaronCrawfis (changes requested):
Please make sure there are no `aliases` in the front-matter of any pages. I saw a couple in the 1.1 branch of the docs-zh repo.

newbe36524:
@AaronCrawfis  I wonder that why it is need to be done those are copied from source file.

Do you think it is ok if I just translated aliases into a url with /zh-hans/ prefix rather than delete them?
You know, to delete them will be manually work for us but to translate them will be easy.

Could you please give me more details about this issue and I can sharing it to my partners.

AaronCrawfis:
> @AaronCrawfis I wonder that why it is need to be done those are copied from source file.
> 
> Do you think it is ok if I just translated aliases into a url with /zh-hans/ prefix rather than delete them?
> You know, to delete them will be manually work for us but to translate them will be easy.
> 
> Could you please give me more details about this issue and I can sharing it to my partners.

@newbe36524 aliases cause issues in Hugo for translations. We add aliases to the English docs anytime we move documentation from one location to another, so that outside bookmarks and links don't break. It acts as a redirect, where if someone navigates to the alias they will instead be redirected to the new path.

If there is an alias in both the English and Chinese docs, Hugo for some reason picks the Chinese version to redirect to. For example, if there is an alias `/my-alias/path`, and a user navigates to https://docs.dapr.io/my-alias/path, they will instead be redirected to the Chinese docs. Original issue here: #1336 

Since Chinese users wouldn't have bookmarks to old versions of the docs, ideally we would delete the aliases in the Chinese translations. However, since it involves work translating the aliases may work. We would need to test that Chinese characters in aliases wouldn't break Hugo or the site since they're translated into URLs. We also need to make sure pages that haven't been translated yet don't contain aliases.

AaronCrawfis:
More information here: https://github.com/gohugoio/hugo/issues/2452

newbe36524:
@AaronCrawfis Thank you very much for being able to provide such detailed information.
I have synced the solution to this issue with the team members to all of them.
In the future, we will prefix these alias with /zh-hans/ to make sure there is no conflict with the en content.
We have verified that this works.
The latest content has been revised up, if you have any questions then feel free to ask.

AaronCrawfis (approved):

PR #1428, opened 2021-04-16
Adding Azure Service Bus publish retry settings

### LINKED ISSUE #1411: [CONTENT] Azure Service Bus: New settings for publishing retry
PR dapr/components-contrib#821 adds the settings `publishMaxRetries` and `publishInitialRetryInternalInMs` in order to control the retry/backoff of publishing to Azure Service Bus when the server is too busy and throttles the connection.

These settings should be added to the ASB pub sub component docs page.

pkedy:
## Description

Adds the settings `publishMaxRetries` and `publishInitialRetryInternalInMs` in order to control the retry/backoff of publishing to Azure Service Bus when the server is too busy and throttles the connection.

Also fixed a few typos where "Event Hubs" was referenced instead of "Service Bus".

## Issue reference

Resolves #1411

### DIFF
--- a/daprdocs/content/en/reference/components-reference/supported-pubsub/setup-azure-servicebus.md
+++ b/daprdocs/content/en/reference/components-reference/supported-pubsub/setup-azure-servicebus.md
@@ -8,7 +8,7 @@ aliases:
 ---
 
 ## Component format
-To setup Azure Event Hubs pubsub create a component of type `pubsub.azure.servicebus`. See [this guide]({{< ref "howto-publish-subscribe.md#step-1-setup-the-pubsub-component" >}}) on how to create and apply a pubsub configuration.
+To setup Azure Service Bus pubsub create a component of type `pubsub.azure.servicebus`. See [this guide]({{< ref "howto-publish-subscribe.md#step-1-setup-the-pubsub-component" >}}) on how to create and apply a pubsub configuration.
 
 ```yaml
 apiVersion: dapr.io/v1alpha1
@@ -50,6 +50,10 @@ spec:
     value: 30
   - name: connectionRecoveryInSec # Optional
     value: 2
+  - name: publishMaxRetries # Optional
+    value: 5
+  - name: publishInitialRetryInternalInMs # Optional
+    value: 500
 ```
 
 > __NOTE:__ The above settings are shared across all topics that use this component.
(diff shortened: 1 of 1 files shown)


daprdocs/content/en/reference/components-reference/supported-pubsub/setup-azure-servicebus.md
```diff
@@ -77,6 +81,8 @@ The above example uses secrets as plain strings. It is recommended to use a secr
 | autoDeleteOnIdleInSec | N  |Time in seconds to wait before auto deleting messages. | `10`
 | maxReconnectionAttempts | N  |Defines the maximum number of reconnect attempts. Default: `30` | `30`
 | connectionRecoveryInSec | N  |Time in seconds to wait between connection recovery attempts. Defaults: `2` | `2`
+| publishMaxRetries | N  | The max number of retries for when Azure Service Bus responds with "too busy" in order to throttle messages. Defaults: `5` | `5`
```
orizohar, line 84: The description implies this only happens for throttling, is that the case? my understanding is that at least right now we don't really know the cause of an error that requires a retry (so this really applies to any retry flow even if not started due to throttling)

daprdocs/content/en/reference/components-reference/supported-pubsub/setup-azure-servicebus.md
```diff
@@ -77,6 +81,8 @@ The above example uses secrets as plain strings. It is recommended to use a secr
...
+| publishInitialRetryInternalInMs | N  | Time in milliseconds for the initial exponential backoff when Azure Service Bus throttle messages. Defaults: `500` | `500`
```
orizohar, line 85: Same comment as above

daprdocs/content/en/reference/components-reference/supported-pubsub/setup-azure-servicebus.md
pkedy: We only retry when we get an error from the ASB SDK indicating the server is "too busy". We made the decision to be specific in this case.

orizohar (approved):

orizohar:
Will merge once release is out

AaronCrawfis (approved):

PR #1471, opened 2021-05-10
add section for api allow lists

### LINKED ISSUE #1460: Allow choosing which APIs are available on a sidecar
This it the docs issue to track feature  https://github.com/dapr/dapr/issues/3054 .

In certain cases, users may want to choose which Dapr APIs are allowed access.
This might be used in low/no-trust environments and reduce an attack surface on a specific API.
In addition, this would allow operators and/or infosec teams to pursue an "enable what you use" philosophy, and comes handy in scenarios where an Ingress might expose the Dapr APIs to external consumers.

yaron2:
Closes https://github.com/dapr/docs/issues/1460.


yaron2:
Wrong branch

PR #1473, opened 2021-05-11
Create api-allowlist.md

### LINKED ISSUE #1460: Allow choosing which APIs are available on a sidecar

yaron2:
Closes #1460.

### DIFF
--- a/daprdocs/content/en/operations/configuration/api-allowlist.md
+++ b/daprdocs/content/en/operations/configuration/api-allowlist.md
@@ -1,0 +1,79 @@
+---
+type: docs
+title: "How-To: Selectively enable Dapr APIs on the Dapr sidecar"
+linkTitle: "Dapr APIs allow list"
+weight: 5000
+description: "Choose which Dapr sidecar APIs are available to the app"
+---
+
+In certain scenarios such as zero trust networks or when exposing the Dapr sidecar to external traffic through a frontend, it's recommended to only enable the Dapr sidecar APIs that are being used by the app. Doing so reduces the attack surface and helps keep the Dapr APIs scoped to the actual needs of the application.
+
+Dapr allows developers to control which APIs are accessible to the application by setting an API allow list using a [Dapr Configuration](({{<ref "configuration-overview.md">}})).
+
+### Default behavior
+
+If an API allow list section is not specified, the default behavior is to allow access to all Dapr APIs.
+Once an allow list is set, only the specified APIs are accessible.
+
+For example, the following configuration enables all APIs for both HTTP and gRPC:
+
+```yaml
+apiVersion: dapr.io/v1alpha1
+kind: Configuration
+metadata:
+  name: myappconfig
+  namespace: default
+spec:
+  tracing:
+    samplingRate: "1"
+```
+
+### Enabling specific HTTP APIs
+
+The following example enables the state `v1.0` HTTP API and block all the rest:
+
+```yaml
+apiVersion: dapr.io/v1alpha1
+kind: Configuration
+metadata:
+  name: myappconfig
+  namespace: default
+spec:
+  api:
+    allowed:
+    - name: state
+      version: v1.0
+```
+
+### Enabling specific gRPC APIs
+
+The following example enables the state `v1` gRPC API and block all the rest:
+
+```yaml
+apiVersion: dapr.io/v1alpha1
+kind: Configuration
+metadata:
+  name: myappconfig
+  namespace: default
+spec:
+  api:
+    allowed:
+    - name: state
+      version: v1
+```
+
+### List of Dapr APIs
+
+The `name` field takes the name of the Dapr API you would like to enable.
+
+See this list of values corresponding to the different Dapr APIs:
+
+| Name  | Dapr API |
+| ------------- | ------------- |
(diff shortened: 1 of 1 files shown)


daprdocs/content/en/operations/configuration/api-allowlist.md
```diff
@@ -0,0 +1,79 @@
...
+      version: v1
+```
+
+### List of Dapr APIs
+
+The `name` field takes the name of the Dapr API you would like to enable.
+
+See this list of values corresponding to the different Dapr APIs:
+
+| Name  | Dapr API |
```
mukundansundar, line 73: @yaron2 Add corresponding version numbers supported and that can be used.

daprdocs/content/en/operations/configuration/api-allowlist.md
```diff
@@ -0,0 +1,79 @@
+---
+type: docs
+title: "How-To: Selectively enable Dapr APIs on the Dapr sidecar"
+linkTitle: "Dapr APIs allow list"
+weight: 5000
```
AaronCrawfis:
```suggestion
weight: 7000
```

daprdocs/content/en/operations/configuration/api-allowlist.md
```diff
@@ -0,0 +1,79 @@
...
+
+| Name  | Dapr API |
+| ------------- | ------------- |
+| state  | State  |
+| invoke  | Service Invocation  |
+| secrets  | Secrets  |
+| bindings  | Output Bindings  |
+| publish | Pub/Sub |
+| actors | Actors |
+| metadata | Metadata |
```
AaronCrawfis:
```suggestion
| state  | [State]({{< ref state_api.md>}})|
| invoke  | [Service Invocation]({{< ref service_invocation_api.md >}})  |
| secrets  | [Secrets]({{< ref secrets_api.md >}})|
| bindings  | [Output Bindings]({{< ref bindings_api.md >}})  |
| publish | [Pub/Sub]({{< ref pubsub.md >}}) |
| actors | [Actors]({{< ref actors_api.md >}}) |
| metadata | [Metadata]({{< ref metadata_api.md >}}) |
```

daprdocs/content/en/operations/configuration/api-allowlist.md
```diff
@@ -0,0 +1,79 @@
...
+apiVersion: dapr.io/v1alpha1
+kind: Configuration
+metadata:
+  name: myappconfig
+  namespace: default
+spec:
+  api:
+    allowed:
+    - name: state
+      version: v1
```
AaronCrawfis, line 63: Is the difference between HTTP and gRPC just the `.0` at the end of the version?

AaronCrawfis (changes requested):

daprdocs/content/en/operations/configuration/api-allowlist.md
yaron2: Yes.

daprdocs/content/en/operations/configuration/api-allowlist.md
yaron2: All versions will be supported going forward.

daprdocs/content/en/operations/configuration/api-allowlist.md
AaronCrawfis: That seems like a weird implementation to choose HTTP vs. gRPC. Can it not be explicitly set? Something like:
```yml
spec:
  api:
    allowed:
    - name: state
      version: v1.0
      protocol: http
```

yaron2: You can't use v1.0 for gRPC APIs because that's not what the protocol versioning is.. a proto version is `v1`, ie only major versions. In that case, the `protocol` field is redundant.

msfussell: We should be explicit here between the HTTP and the gRPC API. Inferring this via the version scheme is not clear

AaronCrawfis: Full example:
```yml
spec:
  api:
    allowed:
    - name: state
      version: v1.0
      protocol: http
    - name: state
      version: v1
      protocol: grpc
```

AaronCrawfis: Inferring grpc from the fact that you choose v1 is only redundant if you have existing knowledge of the versioning scheme of Dapr and of gRPC conventions. I also don't like inferring behavior from another field. The version field should only affect version, not version and protocol.

yaron2: Ok, I agree it looks cleaner. Made the change, please review.

AaronCrawfis (approved):

PR #1564, opened 2021-06-15
Ignore intellij link that isn't resolvable

### LINKED ISSUE #1561: validate-links workflow fails for URL
The validate links workflow is getting a 403 Forbidden error for https://intellij-support.jetbrains.com/hc/en-us/articles/206544519-Directories-used-by-the-IDE-to-store-settings-caches-plugins-and-logs

However it works correctly when manually visited. The site must be blocking scripts and/or GitHub actions.

Script should be updated to fix 403 issue, or create an allow list to pass specific URLs that return a 403 (but still fail for other error codes).

AaronCrawfis:
Ignore https://intellij-support.jetbrains.com/hc/en-us/articles/206544519-Directories-used-by-the-IDE-to-store-settings-caches-plugins-and-logs which can't be resolved

### DIFF
--- a/daprdocs/content/en/developing-applications/ides/intellij.md
+++ b/daprdocs/content/en/developing-applications/ides/intellij.md
@@ -146,4 +146,8 @@ Happy debugging!
 
 ## Related links
 
+<!-- IGNORE_LINKS -->
+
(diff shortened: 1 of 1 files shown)


AaronCrawfis:
https://github.com/dapr/mechanical-markdown/tree/main/examples#link-validation

orizohar (approved):

PR #1596, opened 2021-06-25
Add shortcode to embed sections of code files

AaronCrawfis:
Adds a new `code-snippet` shortcode that:

- Takes as input a file path (relative to /static/code) and a language
- Embeds that code in the webpage
- Optionally a user can specify a `marker` to only embed code between the marker
- Optionally a user can specify a find/replace for sections of code that are too verbose and don't need to be rendered in docs.

This unlocks SDK code to be mounted into the /static/code folder and referenced via this shortcode.


daprdocs/content/en/contributing/contributing-docs.md
```diff
@@ -219,6 +219,102 @@ brew install dapr/tap/dapr-cli
...
+
+        time.sleep(2)
+#SAMPLE
+```
+
+Will result in the following output:
+
+{{< code-snippet file="contributing-1.py" lang="python" marker="#SAMPLE" >}}
+
+Use the `replace-key-[token]` and `replace-value-[token]` parameters to limit the embedded snipped to a portion of the sample file. This is useful when you want abbreviate a portion of the code sample. Multiple replacements are supported with multiple values of `token`.
```
msfussell, line 277: What is the difference between this and the #SAMPLE above? They achieve the same thing, yes?

msfussell (approved):
Very nice addition!

PR #1604, opened 2021-07-01
Comment out all optional destructive parameters on Azure Service Bus

AaronCrawfis:


msfussell:
We can do this to SB config. I suspect that we need to also apply this to all other optional parameters on all components.

msfussell (approved):

PR #1623, opened 2021-07-07
Add doc for blob storage metadata retrieval and blob listing

akkie:
## Description

Add documentation for the retrieval of user defined blob storage metadata.

## Issue reference

See: https://github.com/dapr/components-contrib/issues/683


AaronCrawfis (approved):

AaronCrawfis (approved):

yaron2:
@AaronCrawfis this can be merged. Reference: https://github.com/dapr/components-contrib/pull/999

akkie:
@AaronCrawfis Would it be OK if I fix the Zeebe link in this PR?

orizohar:
> @AaronCrawfis Would it be OK if I fix the Zeebe link in this PR?

Thanks for suggesting! We fixed it in v1.2 and will do an upmerge soon that will also fix it in v1.3

PR #1678, opened 2021-07-21
Add a list of all current preview features

### LINKED ISSUE #1592: Need topic that list all the **preview features** and when they were introduced
Each release can now have a list of preview features. Under Support there needs to be a table that list the preview features, which build they were introduced in, a link to the topic and which build they became GA in. This way there is a single table in the docs updated after each release, that clearly provides the latest information, rather than searching through multiple docs and having to update them.

halspang:
## Description

This commits provides a single point of reference for all
currently supported preview features.

## Issue reference

This PR will close: #1592


halspang:
Note that we need the other preview feature docs to merge before this is OK.

orizohar (approved):

AaronCrawfis (approved):

PR #1730, opened 2021-08-19
Describe the use of unix sockets, the CLI commands and usage example.md

daixiang0:
Ref https://github.com/dapr/dapr/pull/3529


orizohar:
@daixiang0 , @artursouza  - is this new option only for daprd? was there no change in the CLI that exposes this to users as well? if so, we need to add it in the CLI column as well...

daixiang0:
yep, there it is https://github.com/dapr/cli/pull/788

orizohar:
@artursouza, @wcs1only - so is this going to only be in daprd for v1.4 and then added as a CLI option in v1.5? if the CLI also has an added option in v1.4 we need to add it to this table as well.

artursouza:
CLI will be done later in 1.5. We might even need to update the daprd param to make it accept a string and not just a boolean.

orizohar:
@daixiang0 from the discussion on https://github.com/dapr/dapr/pull/3529 looks like this should go into v1.5 - please change the PR to merge to the v1.5 branch of the docs

daixiang0:
@orizohar done.

willtsai:
@daixiang0 - besides the additional row you've added in arguments-annotations-overview.md, is there another place in Dapr docs where you can add more explanation for this new feature? We need to make sure users get more details about when to use this feature, what it's used for, etc.

msfussell:
Adding to @willtsai request. There really needs to be some explaination in the docs as to why someone would use this feature and more how. That means creating a new doc for this. At least include words and a description here and then we can determine a best place for this. 

And adding an example like you have in the code PR to explain usage. like this.

run with hello-world example in quick-start
daprd --app-id nodeapp --app-protocol http --components-path /root/.dapr/components --metrics-port 39497 --dapr-http-max-request-size -1 --app-port 3000 --placement-host-address localhost:50005 --config /root/.dapr/config.yaml --enable-domain-socket
test
curl --unix-socket /tmp/nodeapp-http.socket http://unix/v1.0/state/statestore/order

msfussell:
'run with hello-world example in quick-start
daprd --app-id nodeapp --app-protocol http --components-path /root/.dapr/components --metrics-port 39497 --dapr-http-max-request-size -1 --app-port 3000 --placement-host-address localhost:50005 --config /root/.dapr/config.yaml --enable-domain-socket
test
curl --unix-socket /tmp/nodeapp-http.socket http://unix/v1.0/state/statestore/order

daixiang0:
@msfussell @willtsai I would do that when codes are stable based on reviews.

willtsai:
@daixiang0 - thank you. however, can you please elaborate on what other code needs to be more stable based on reviews? My understanding is that the features mentioned in this docs PR has already been merged as a part of https://github.com/dapr/dapr/pull/3529 and thus we need to make sure the documentation is updated to reflect that code change.

daixiang0:
What does the "stable" mean?

willtsai:
@daixiang0 - I was referencing your previous comment where you mention "stable" :)
"I would do that when codes are stable based on reviews."

daixiang0:
@willtsai done.

daprdocs/content/en/reference/arguments-annotations-overview.md
```diff
@@ -27,6 +27,7 @@ This table is meant to help users understand the equivalent options for running
 | `--enable-metrics` | not supported | | configuration spec | Enable prometheus metric (default true) |
 | `--enable-mtls` | not supported | | configuration spec | Enables automatic mTLS for daprd to daprd communication channels |
 | `--enable-profiling` | `--enable-profiling` | | `dapr.io/enable-profiling` | Enable profiling |
+| `--enable-domain-socket` | not supported | | not supported  | Enable listen with Unix domain socket |
```
greenie-msft:
```suggestion
| `--enable-domain-socket` | not supported | | not supported  | Enables listening with unix domain sockets for lower latency and greater throughput when an app calls into the Dapr sidecar.  Not available on Windows OS. |
```

daprdocs/content/en/reference/cli/dapr-run.md
```diff
@@ -38,6 +38,7 @@ dapr run [flags] [command]
 | `--log-level`                  |                      | `info`                                                                             | The log verbosity. Valid values are: `debug`, `info`, `warn`, `error`, `fatal`, or `panic`           |
 | `--metrics-port`               | `DAPR_METRICS_PORT`  | `9090`                                                                             | The port that Dapr sends its metrics information to                                                  |
 | `--profile-port`               |                      | `7777`                                                                             | The port for the profile server to listen on                                                         |
+| `--unix-domain-socket`, `-u`   |                      |                                                                                    |  Path to a unix domain socket dir mount. If specified, Dapr API servers will use Unix Domain Sockets |
```
greenie-msft:
```suggestion
| `--unix-domain-socket`, `-u`   |                      |                                                                                    |  Path to a unix domain socket dir mount. If specified, Dapr API servers will use unix domain sockets for lower latency and greater throughput when an app calls into the Dapr sidecar. Not available on Windows OS. |
```

daprdocs/content/en/reference/cli/dapr-run.md
```diff
@@ -46,6 +47,9 @@ dapr run [flags] [command]
 # Run a .NET application
 dapr run --app-id myapp --app-port 5000 -- dotnet run
 
+# Run a .Net application with socket
```
greenie-msft:
```suggestion
# Run a .Net application with unix domain sockets
```

daixiang0: Fixed.

daprdocs/content/en/reference/arguments-annotations-overview.md
daixiang0: Done.

daprdocs/content/en/reference/cli/dapr-run.md
daixiang0: done.

willtsai (approved):

greenie-msft (approved):

PR #1732, opened 2021-08-19
clean up all whitenoise

daixiang0:
Signed-off-by: Long <long0dai@foxmail.com>


greenie-msft (approved):

orizohar:
@daixiang0 - This PR is not a necessary one since whitespaces do not effect the rendering of the website. We do not enforce this in docs PRs so I imagine this will continue to show up. Doesn't mean this PR won't be merged but just I think the continuous effort is not necessary in the future

orizohar (approved):

PR #1734, opened 2021-08-20
Mentioning multi-valued secret support for local file secret store

### LINKED ISSUE #1733: Mention multiValue support metadata for local file secret store.
Mention multiValue support metadata for local file secret store.

Mention new feature.

https://docs.dapr.io/reference/components-reference/supported-secret-stores/file-secret-store/

https://github.com/dapr/components-contrib/pull/1089

pkedy:
## Description

Adding the documentation/setting for multi-valued secrets in the local file secret store.

Merged PR dapr/components-contrib#1089

## Issue reference

Closes #1733

/cc @artursouza


daprdocs/content/en/reference/components-reference/supported-secret-stores/file-secret-store.md
```diff
@@ -54,15 +57,32 @@ Given the following json:
...
-Use the flattened key (`connectionStrings:sql`) to access the secret.
+Use the flattened key (`connectionStrings:sql`) to access the secret. The following JSON map returned:
+
+```json
+{
+  "connectionStrings:sql": "your sql connection string"
+}
+```
+
+If `multiValued` is `"true"`, you would instead use the top level key. In this example, `connectionStrings` would return the following map:
```
artursouza, line 76: Please, also mention how the subtrees are flat here as well and appended with the separator. It was the main difference between your approach and the one I had in mind. I actually think your approach is better and should be documented here for user's benefit.

artursouza (changes requested):

daprdocs/content/en/reference/components-reference/supported-secret-stores/file-secret-store.md
pkedy: @artursouza See new content below

artursouza (approved):

orizohar (approved):

PR #1773, opened 2021-09-09
Create variable for latest Dapr version and update all version refere…

greenie-msft:
…nces to useit

## Description

Created a shortcode "dapr-version". Updated all latest version references to use the shortcode.


daprdocs/layouts/shortcodes/dapr-version.html
```diff
@@ -0,0 +1 @@
+1.4.0
```
willtsai: should this file (and consequently its shortcodes) be named "dapr-latest-released-version" to be more explicit?

greenie-msft: I can do that.

greenie-msft:
Fwiw, the validations are failing because the quickstarts have not yet cut v1.4 versions.

daprdocs/content/en/getting-started/quickstarts.md
```diff
@@ -17,11 +17,11 @@ The [Dapr Quickstarts](https://github.com/dapr/quickstarts/tree/v1.0.0) are a co
...
-| [Bindings](https://github.com/dapr/quickstarts/tree/v1.3.0/bindings)            | Demonstrates how to use Dapr to create input and output bindings to other components. Uses bindings to Kafka.                                                                            |
-| [Middleware](https://github.com/dapr/quickstarts/tree/v1.3.0/middleware) | Demonstrates use of Dapr middleware to enable OAuth 2.0 authorization. |
-| [Observability](https://github.com/dapr/quickstarts/tree/v1.3.0/observability) | Demonstrates Dapr tracing capabilities. Uses Zipkin as a tracing component. |
-| [Secret Store](https://github.com/dapr/quickstarts/tree/v1.3.0/secretstore) | Demonstrates the use of Dapr Secrets API to access secret stores. |
+| [Hello World](https://github.com/dapr/quickstarts/tree/v{{% dapr-latest-version %}}/hello-world)            | Demonstrates how to run Dapr locally. Highlights service invocation and state management.                                                                                                      |
```
orizohar: For all of these, the link validation fails, need to have the validation script ignore these

greenie-msft: I would think we would just wait until those links are available? I don't think we should ignore these links in general.

daprdocs/layouts/shortcodes/dapr-latest-version.html
```diff
@@ -0,0 +1 @@
+1.4.0
```
orizohar: Since some places expect 1.4 and some 1.4.0 I think this should be 1.4 and the where needed add the suffix .0

daprdocs/content/en/getting-started/quickstarts.md
orizohar: I don't think that would work. The validator doesn't replace the var, it's a script separate from Hugo

daprdocs/layouts/shortcodes/dapr-latest-version.html
greenie-msft: I actually created a conditional in the short-code to handle this based on params being set. It's cleaner than leaving .0's around the markdown.

daprdocs/layouts/shortcodes/dapr-latest-version.html
```diff
@@ -0,0 +1,5 @@
+{{ if .Get "version" }}
+1.4
+{{ else }}
+1.4.0
+{{ end }}
```
willtsai: Maybe a nit, but I would prefer the implementation to be more explicit to deal with when the shortcode specifies either version="short" or version="long" to avoid someone in the future using something like ``{{% dapr-latest-version version="long" %}}`` and getting unexpected results. The ``else`` to default to the long version makes sense though.

orizohar:
Once all comments are addressed, this doesn't really need to wait for the "day-of" to be merged

willtsai (approved):
LGTM! thanks for finding a way to make the version short/long tag explicit.

daprdocs/layouts/shortcodes/dapr-latest-version.html
greenie-msft: Done.

greenie-msft:
@orizohar I don't think the ignore_links works inside the table. Tried multiple approaches with no luck.

orizohar:
Looks like we'll need to keep the quicstart links updated manually. Bringing back the do-not-merge label until quickstarts cut the release branch

greenie-msft:
@orizohar I've updated the quickstart links (manually).

orizohar (approved):

PR #1819, opened 2021-09-30
Disallow iFrame embeds

berndverst:
Disallow iFrame embeds

### DIFF
--- a/daprdocs/config.toml
+++ b/daprdocs/config.toml
@@ -10,6 +10,12 @@ enableGitInfo = true
 # Language Configuration
 languageCode = "en-us"
 
+[server]
+[[server.headers]]
+  for = '/**.html'
+  [server.headers.values]
+    X-Frame-Options = 'DENY'
+
 [languages]
   [languages.en]
     title = "Dapr Docs"


artursouza (approved):

daprdocs/config.toml
```diff
@@ -10,6 +10,12 @@ enableGitInfo = true
 # Language Configuration
 languageCode = "en-us"
 
+[server]
+[[server.headers]]
+  for = '/**.html'
```
pkedy, line 15: In my local testing, `/**.html` did not work. I had to use `/**` for the response headers to return. However, when I did that, Chrome developer extensions showed something interesting for `.js` files.  I'm not sure its its a real issue or not.

image

pkedy: My config snippet was:

```toml
[server]
[[server.headers]]
  for = '/**'
  [server.headers.values]
    Content-Security-Policy = 'script-src localhost:1313'
    Referrer-Policy = 'strict-origin-when-cross-origin'
    X-Content-Type-Options = 'nosniff'
    X-Frame-Options = 'DENY'
    X-XSS-Protection = '1; mode=block'
```

AaronCrawfis:
[server] works for local dev only. For production we'll need to define a Static Web Apps configuration file and update the global headers: https://docs.microsoft.com/en-us/azure/static-web-apps/configuration#global-headers

berndverst:
See #1820 @pkedy  @AaronCrawfis

berndverst:
I'm closing this because it only applies to local development and we don't really need this protection there.

PR #1846, opened 2021-10-08
Update cosmosDB state `partitionkey` usage

### LINKED ISSUE #1845: Cosmos DB state store `partitionKey` example is invalid for custom partitions
**Describe the issue**
Cosmos DB state store documentation on specifying [partition key values](https://docs.dapr.io/reference/components-reference/supported-state-stores/setup-azure-cosmosdb/#partition-keys) is incorrect when a user tries to use it for custom defined partition keys (not just the partition key value), and does not accurately convey how the property can be successfully used.

See https://github.com/dapr/cli/issues/805 for details.

**URL of the docs**
https://docs.dapr.io/reference/components-reference/supported-state-stores/setup-azure-cosmosdb/#partition-keys

**Expected content**
The doc should provide a usable example of setting the `partitionKey` value, in relation to how the target cosmosDB container partition key was configured.

See https://github.com/dapr/cli/issues/805 for details.

**Additional context**
There is an additional issue that may be a functional bug, where the default expectation that the `key` property can be used by the state store when `partitionKey` is not specified is effectively unusable because the component always modifies the `key` value to create the `id` in the cosmosDB entry, and then expects the `id` value to be the partition key value which would not occur naturally.

CodeMonkeyLeet:

## Description

- Increase emphasis of default `/paritionKey` partition key for CosmosDB collections for use with Dapr state stores.
- Add example of using `partitionKey` metadata with CosmosDB collection using custom partition keys.
- Add `keyPrefix` configuration documentation to state_api.md.

## Issue reference

Fixes #1845

### DIFF
--- a/daprdocs/content/en/reference/api/state_api.md
+++ b/daprdocs/content/en/reference/api/state_api.md
@@ -403,12 +403,22 @@ spec:
 
 ## Optional behaviors
 
-### Key scheme
+### Custom key schemes
 
-A Dapr-compatible state store shall use the following key scheme:
+For non-actor state stores, the key scheme can be customized by setting the `keyPrefix` metadata property in the component's configuration YAML file. The `keyPrefix` can be set to one of the following values:
 
-* *\<App ID>||\<state key>* key format for general states
-* *\<App ID>||\<Actor type>||\<Actor id>||\<state key>* key format for Actor states.
+| `keyPrefix` value | Key scheme | Example |
+| ----------------- | ---------- | ------- |
+| `appid` (default) | _\<app ID\>_\|\|_\<state key\>_        | `myapp\|\|key1` |
+| `name`            | _\<store name\>_\|\|_\<state key\>_    | `myredisstore\|\|key1` |
+| `none`            | _\<state key\>_                        | `key1` |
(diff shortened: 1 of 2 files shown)


CodeMonkeyLeet:
@artursouza @berndverst to review for technical accuracy. The behaviors documented are empirically determined, but it's unclear to me whether all of it is intended.

daprdocs/content/en/reference/api/state_api.md
```diff
@@ -403,12 +403,22 @@ spec:
 
 ## Optional behaviors
 
-### Key scheme
```
artursouza, line 406: Keep the title.

artursouza (changes requested):
The Dapr state store API is an opinionated implementation and does not offer a "bring your own schema" functionality. For that, we plan to support a new API for databases, where apps will be able to bring their own schemas and key structure.

https://github.com/dapr/dapr/issues/3354

daprdocs/content/en/reference/api/state_api.md
CodeMonkeyLeet: @artursouza FYI, the key scheme section under Optional Behaviors is redundant as there is a top level `## Key Scheme` section on line 37 that repeats the same info. Given that the key scheme isn't really an optional behavior, it should probably be removed.

PR #1902, opened 2021-11-04
SQL Server IndexedProperties example

berndverst:
Provides correct example of SQL Server IndexedProperties

The existing example was syntactically wrong and could never have worked. This makes me question whether anyone has ever used this feature before.


berndverst:
@willtsai  @greenie-msft

willtsai (approved):

greenie-msft (approved):

PR #1929, opened 2021-11-11
Correct metadata key name in Azure Key Vault secret store

ItalyPaleAle:
`spn*` are legacy keys


willtsai:
this looks good as I see the keys referenced here: https://github.com/dapr/components-contrib/blob/master/authentication/azure/metadata-properties.go

however, @ItalyPaleAle - can you please change the base merge branch to v1.5 since that's the most recently released version of Dapr?

ItalyPaleAle:
> can you please change the base merge branch to v1.5 since that's the most recently released version of Dapr?

Yeah I created this PR at a bad time, during the endgame. Because changing the base messed up everything, I have re-created the PR in #1985. Will close this

> this looks good as I see the keys referenced here: https://github.com/dapr/components-contrib/blob/master/authentication/azure/metadata-properties.go

Yes, I wrote that code :)

PR #1985, opened 2021-11-19
 Correct metadata key name in Azure Key Vault secret store 

ItalyPaleAle:
`spn*` are legacy keys
Re-do of #1929 with a new base


ItalyPaleAle:
@orizohar @willtsai can you please re-open and merge this? It's a quick fix

PR #2051, opened 2021-12-17
Fix pub/sub subscription header

AaronCrawfis:
Somehow we lost a header with subscriptions


AaronCrawfis:
Actually I misread the code. We didn't mess up a header, we just ommited the programmatic section. Will close this PR for now

PR #2155, opened 2022-01-26
Getting Started restructure and Pub/sub quickstart

hhunter-ms:
## Description

- New pub/sub quickstart
- Reorganize Getting Started area for better accessibility
- Freshness grammar/style pass on docs w/i Getting Started area

### Need

- repo used for `git clone` in pub/sub quickstart

## Issue reference

#2213


.vscode/settings.json
```diff
@@ -0,0 +1,8 @@
+{
```
greenie-msft, line 1: We should exclude this file from the PR

daprdocs/content/en/getting-started/quickstarts/pubsub-quickstart.md
```diff
@@ -0,0 +1,605 @@
...
+description: "Get started with Dapr's Publish and Subscribe building block"
+---
+
+Let's take a look at Dapr's [Publish and Subscribe (Pub/sub) building block]({{< ref pubsub >}}). In this quickstart, you will set up a publisher microservice and a subscriber microservice to demonstrate how Dapr enables a Pub/sub pattern.
+
+1. Using a publisher service, developers can repeatedly publish messages to a topic.
+1. [A Pub/sub component](https://docs.dapr.io/concepts/components-concept/#pubsub-brokers) queues or brokers those messages (using Redis Streams, RabbitMQ, Kafka, etc.).
+1. The subscriber to that topic pulls messages from the queue and processes them.
+
+Select your preferred language before proceeding with the quickstart.
```
greenie-msft: Consider being explicit that these code tabs are actually showing the examples using Dapr SDKs that are language specific.

daprdocs/content/en/getting-started/quickstarts/pubsub-quickstart.md
```diff
@@ -0,0 +1,605 @@
...
+
+Or:
+
+```bash
+python -m pip install -r requirements.txt
+```
+
+### Step 2: View the Pub/sub component
+
+When you run `dapr init`, Dapr creates a default redis `pubsub.yaml` on your local machine. Verify by opening your components directory:
```
greenie-msft: It also runs Redis in a container on  your local machine. That's what the component is connecting to.

daprdocs/content/en/getting-started/quickstarts/pubsub-quickstart.md
```diff
@@ -0,0 +1,605 @@
...
+ <!-- Python -->
+{{% codetab %}}
+
+### Pre-requisites
+
+For this example, you will need:
+
+- [Dapr CLI and initialized environment](https://docs.dapr.io/getting-started).
+- [Python 3.7+ installed](https://www.python.org/downloads/).
+
```
greenie-msft, line 28: For all quickstarts, Docker Desktop is also a pre-req since we're relying on redis for most examples

daprdocs/content/en/getting-started/quickstarts/pubsub-quickstart.md
```diff
@@ -0,0 +1,605 @@
...
+```
+
+### Step 2: View the Pub/sub component
+
+When you run `dapr init`, Dapr creates a default redis `pubsub.yaml` on your local machine. Verify by opening your components directory:
+
+- On Windows, under `%UserProfile%\.dapr\components\pubsub.yaml`
+- On Linux/MacOS, under `~/.dapr/components/pubsub.yaml`
+
+For this quickstart, we've included a default Redis `pubsub.yaml` file within the cloned repository that contains the following:
```
greenie-msft: This sentence seems a little redundant to the one above.

daprdocs/content/en/getting-started/quickstarts/pubsub-quickstart.md
```diff
@@ -0,0 +1,605 @@
...
+
+{{< tabs "Self-Hosted (CLI)" "Kubernetes" >}}
+
+{{% codetab %}}
+
+```yaml
```
greenie-msft: Consider pointing out what is defined in this component YAML. 

For example, the name is how your application talks to the component. The metadata defines the connection to the instance of the component. Scopes specify which application can use the component.

daprdocs/content/en/getting-started/quickstarts/pubsub-quickstart.md
```diff
@@ -0,0 +1,605 @@
...
+Navigate to the directory containing your subscriber application.
+
+```bash
+cd pub_sub/python
+```
+
+With the following command, run the CheckoutService.py service alongside a Dapr sidecar.
+
+```bash
+dapr run --app-id checkout --app-port 6002 --dapr-http-port 3602 --app-protocol grpc --components-path ../components  python3 CheckoutService.py
```
greenie-msft: if app protocol is grpc, i'm not sure dapr-http-port is required. Instead, I would imagine dapr-grpc-port is needed

daprdocs/content/en/getting-started/quickstarts/pubsub-quickstart.md
```diff
@@ -0,0 +1,605 @@
...
+Navigate to the directory holding the "Order Processing Service" Program.cs publisher.
+
+```bash
+cd pub_sub/csharp/OrderProcessingService
+```
+
+With the following command, run Program.cs service alongside a Dapr sidecar.
+
+```bash
+dapr run --app-id orderprocessing --app-port 6001 --dapr-http-port 3601 --dapr-grpc-port 60001 --app-ssl --components-path ../../components dotnet run
```
greenie-msft: Same here - if app protocol is grpc, i'm not sure dapr-http-port is required. Instead, I would imagine dapr-grpc-port is needed

.vscode/settings.json
paulyuk: Overall good stuff. 
 
I have comments and questions:
 

	Python quickstart is coming along.  I like the changes.  
	Where is the code for these?  I think the example is still backwards where an order processor calls the cart.  
	.NET needs a bit of work.  
	
		The term we should use is simply .NET SDK or .NET 6 SDK.  "Core" is not term we use anymore.  SDK is the specific deliverable vs. the runtime.  the runtime isn't enough.  
		The hard coded download link isn't what we want.  lets use this: https://dotnet.microsoft.com/en-us/download
		I don't understand what we mean by get latest Nuget packages a prereq.  We can remove that.  Instead of doing that, have a step to `dotnet restore` and `dotnet build `  These will bring the nuget package refs back for you, assuming they are already in sample.  This needs to happen before `dapr run` steps.  This is moral equivalent of the `pip install` you did in Python.  Dotnet is a compiled language so that dotnet build also needs to happen, not just the restore.  
		The code is inside a `main` function loop.  In .NET 6 we don't do that anymore.  You just start coding in the Program.cs without a main.  See this example: [ASPNETCoreFrontEndWithBackendAPIsonContainerApps/Program.cs at paulyuk/dapr · paulyuk/ASPNETCoreFrontEndWithBackendAPIsonContainerApps (github.com)](https://github.com/paulyuk/ASPNETCoreFrontEndWithBackendAPIsonContainerApps/blob/paulyuk/dapr/DotNetOnContainerApps.FrontEnd/Program.cs)
		Also for the snip of code we just want to show the publish or subscribe part, not all the imports and sleeps, etc.
	
	
	Javascript
	
		instead of npm install on daprclient, should we just `npm install` and let the package list take care of it?
		do we need an npm run buildclient ?
		should we do apples to apples with python and remove the function wrapping the publish code snippet?
	
	
	deploy to cloud -- in a future rev (dont block on it now) i'd like to point to a cloud integrations page that show how to deploy to each cloud.  We could write the azure one and invite community to do the others

paulyuk (changes requested):
Please see comments above for suggested updates.  Nice progress.  Thank you.

hhunter-ms:
Closing since PR is out of date (against v1.5). Will open new PR against v1.6.

PR #2168, opened 2022-01-30
Update overview.md

qt-luigi:
Fix typo. (one word)


qt-luigi:
Sorry, this is a local.

PR #2194, opened 2022-02-10
Update JS sdk submodule which will bring in doc changes to use http a…

greenie-msft:
Update JS sdk submodule which will bring in doc changes to use http a…


msfussell (approved):
LGTM

PR #2223, opened 2022-03-02
Update get-started-component.md

duglin:
Teach them about "dapr stop" while we're at it

### DIFF
--- a/daprdocs/content/en/getting-started/get-started-component.md
+++ b/daprdocs/content/en/getting-started/get-started-component.md
@@ -65,7 +65,7 @@ Run the following command to launch a Dapr sidecar that will listen on port 3500
 dapr run --app-id myapp --dapr-http-port 3500 --components-path ./my-components
 ```
 
-> If you encounter a error message stating the app ID is already in use, it may be that the sidecar you ran in the previous step is still running. Make sure you stop the sidecar before running the above command (e.g. using "Control-C").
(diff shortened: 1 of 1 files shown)


daprdocs/content/en/getting-started/get-started-component.md
```diff
@@ -65,7 +65,7 @@ Run the following command to launch a Dapr sidecar that will listen on port 3500
 dapr run --app-id myapp --dapr-http-port 3500 --components-path ./my-components
 ```
 
-> If you encounter a error message stating the app ID is already in use, it may be that the sidecar you ran in the previous step is still running. Make sure you stop the sidecar before running the above command (e.g. using "Control-C").
+> If you encounter a error message stating the app ID is already in use, it may be that the sidecar you ran in the previous step is still running. Make sure you stop the sidecar before running the above command (e.g. using "Control-C" or running the command `dapr stop --app-id myapp`).
```
msfussell: Great comment! really do not need the e.g. in the sentence, just say what you actually need to do. 

```suggestion
> If you encounter a error message stating the app ID is already in use, it may be that the sidecar you ran in the previous step is still running. Make sure you stop the sidecar before running the above command using "Control-C" or running the command `dapr stop --app-id myapp`.
```

duglin: works for me. Updated commit.

msfussell (approved):
LGTM. Thanks for your contribution to the docs!

Issue #2186, opened 2022-02-08
Add information about setting content type for objects in the state store


dmitsh:
**What content needs to be created or modified?**
This is an [old ask](https://github.com/dapr/dapr/issues/2892) to differentiate content types of stored objects in the state store.
This will allow to properly store, retrieve, and query data.

**Describe the solution you'd like**

The solution is described in https://github.com/dapr/dapr/issues/2892, https://github.com/dapr/dapr/pull/4210

**Where should the new material be placed?**

**The associated pull request from dapr/dapr, dapr/components-contrib, or other Dapr code repos**

https://github.com/dapr/dapr/pull/4210

**Additional context**

msfussell:
Resolved.

PR #2225, opened 2022-03-03
how to set content type in state store

dmitsh:

## Description

Added info on how to set content type in state store

## Issue reference

#2186

### DIFF
--- a/daprdocs/content/en/developing-applications/building-blocks/state-management/state-management-overview.md
+++ b/daprdocs/content/en/developing-applications/building-blocks/state-management/state-management-overview.md
@@ -55,6 +55,17 @@ When strong consistency is used, Dapr waits for all replicas (or designated quor
 
 Read the [API reference]({{< ref state_api.md >}}) to learn how to set consistency options.
 
+### Setting content type
+
+Some state store components might have different ways of maintaining and manipulating data, depending on its type.
+Dapr supports passing content type in [State management API](#state-management-api) as part of request metadata.
+
+Setting the content type is optional, and it is up to the component to make use of it. Dapr only provides the means of passing this information to the component.
+
+With the HTTP API, you can set content type via URL query parameter `metadata.contentType`. For example, `http://localhost:3500/v1.0/state/store?metadata.contentType=application/json`
+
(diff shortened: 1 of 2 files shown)


mukundansundar (approved):
LGTM

daprdocs/content/en/developing-applications/building-blocks/state-management/state-management-overview.md
```diff
@@ -55,6 +55,15 @@ When strong consistency is used, Dapr waits for all replicas (or designated quor
 
 Read the [API reference]({{< ref state_api.md >}}) to learn how to set consistency options.
 
+### Setting content type
+
+Some state store components might have different ways of maintaining and manipulating data, depending on its type.
+Dapr supports passing content type in [State management API](#state-management-api) as part of request metadata.
```
msfussell, line 61: I presume that content type is optional. If so, can you say that. And if not supplied what does the contentType default to.

msfussell:
Few other things to fix up in state_api.md file
1) Remove this sentence "Starting with 0.4.0 release, support for multiple state stores was added. This is a breaking change from previous releases as the state APIs were changed to support this new scenario."
2) It is important to describe how to use contentType here in the API reference. Suggest coping some of the words from the overview section, rather than just show an example. 
3) Are there other values than 'metadata.contentType=application/json` that can be mentioned in the description? 
4) Can you also cover whether metadata.contentType is optional. Adding this to all the API calls seems to imply that this is mandatory. And if not supplied, what is the default?
5) Are there any other places in this API spec you feel could be improved with clearer or more examples. This always helps if you have some ideas for improvement.

dmitsh:
> * It is important to describe how to use contentType here in the API reference. Suggest coping some of the words from the overview section, rather than just show an example.
> * Are there other values than 'metadata.contentType=application/json` that can be mentioned in the description?
> * Can you also cover whether metadata.contentType is optional. Adding this to all the API calls seems to imply that this is mandatory. And if not supplied, what is the default?

I would disagree. If we describe a particular metadata in the API reference, then we need to describe all possible metadata as well.
Also, I have already mentioned in the overview that `metadata.contentType` is optional. Dapr does not set the permitted values for the content type. It is all up to the component.

PR #2230, opened 2022-03-04
appID when scaling is static - rewording

duglin:
On: https://docs.dapr.io/developing-applications/building-blocks/service-invocation/service-invocation-overview/
it says:
- - -
Note: You can have N instances of the same app with the same app ID as app ID is unique per app. And you can have multiple instances of that app where all those instances have the same app ID
- - -
I found this paragraph hard to parse. If it's just me go ahead and close this PR, but I tried to capture what I think it's trying to say in more clear language.


msfussell (approved):
LGTM

PR #2231, opened 2022-03-04
Fix alignment in kubectl output

duglin:
Sorry, my AR got the better of me :-)  but:
image
was bugging me


msfussell:
@duglin - Thanks for this contribution. Really this is best to go into the v1.6 branch (current) rather than v1.7 (next) but given this is close we can do this. If there are other updates like this target v1.6 please. Thanks

msfussell (approved):
LGTM

PR #2233, opened 2022-03-04
Point to the nested tracing doc

duglin:
Just a minor help for readers... take them directly to the doc in question
rather than to the TOC, from which they then need to cick to get to where
they want to be.


msfussell:
@duglin - Thanks for this contribution. Really this is best to go into the v1.6 branch (current) rather than v1.7 (next) but given this is close we can do this. If there are other updates like this target v1.6 please. Thanks

msfussell (approved):
LGTM

PR #2241, opened 2022-03-09
Add clarification on authenticating to Azure

AaronCrawfis:
Clarifies example to point further down below


msfussell:
Need to add the DCO!
These changes - Do they align better with the name values in the Azure API? Is that the reason for this update?

AaronCrawfis:
> Need to add the DCO! These changes - Do they align better with the name values in the Azure API? Is that the reason for this update?

DCO added

This PR fixes examples to use non-deprecated values, as well as provide better guidance on how to use auth fields. There was some confusion on our team on how to use Key Vaults.

msfussell (approved):

PR #2246, opened 2022-03-10
Update s3 binding

yaron2:
S3 binding updated with new metadata in https://github.com/dapr/components-contrib/issues/1554.


msfussell (approved):
LGTM

PR #2260, opened 2022-03-14
Clean up building block how-tos

greenie-msft:
This is a draft. Please do not merge.

Cleaning up how-tos by removing the "step" sections which imply that the guides are sequential. 

Also removing dapr run commands since there is no src code associated.


greenie-msft:
PR created against wrong branch. Closing.

PR #2261, opened 2022-03-14
Cleaning up how-tos by removing the "step" sections which imply that …

greenie-msft:
This is a draft. Please do not merge.

Cleaning up how-tos by removing the "step" sections which imply that the guides are sequential.

Also removing dapr run commands since there is no src code associated


hhunter-ms:
closing per discussion with @msfussell . redo of PR w/o conflicts: #2542

PR #2266, opened 2022-03-17
Resiliency Docs [initial version]

berndverst:
Signed-off-by: Bernd Verst <4535280+berndverst@users.noreply.github.com>

These are the initial [incomplete] docs for resiliency.

This PR will need help from @greenie-msft on wording.

Also there are a few questions for @halspang and @pkedy:
- Can you please propose text that explains circuitbreakers and explain the relevant options?
- Sometimes the example uses `general` policies for retry and and circuitbreaking. Are these default policies that come for free with the runtime? Or must every policy be explicitly defined? If the latter the example needs to be updated (and the testdata in our repo is wrong).

### DIFF
--- a/daprdocs/content/en/operations/security/mtls.md
+++ b/daprdocs/content/en/operations/security/mtls.md
@@ -256,6 +256,50 @@ kubectl rollout restart deploy/myapp
 
 You will experience potential downtime due to mismatching certificates until all deployments have successfully been restarted (and hence loaded the new Dapr certificates).
 
+### Set up monitoring for Dapr control plane mTLS certificate expiration
+
+Beginning 30 days prior to mTLS root ceritificate expiration the Dapr sentry service will hourly emit warning level logs indicating that the root certificate is about to expire.
+
+As an operational best practice for running Dapr in production we recommend configuring monitoring for these particular sentry service logs so that you are aware of the upcoming certificate expiration.
+
+```bash
+"Dapr root certificate expiration warning: certificate expires in 2 days and 15 hours"
+```
+
+Once the certificate has expired you will see the following message:
+
+```bash
+"Dapr root certificate expiration warning: certificate has expired."
+```
+
+In Kubernetes you can view the sentry service logs like so:
+
+```bash
+kubectl logs deployment/dapr-sentry -n dapr-system
+```
+
+The log output will appear like the following:"
+
+```bash
+{"instance":"dapr-sentry-68cbf79bb9-gdqdv","level":"warning","msg":"Dapr root certificate expiration warning: certificate expires in 2 days and 15 hours","scope":"dapr.sentry","time":"2022-04-01T23:43:35.931825236Z","type":"log","ver":"1.6.0"}
+```
+
+As an additional tool to alert you to the upcoming certificate expiration beginning with release 1.7.0 the CLI now prints the certificate expiration status whenever you interact with a Kubernetes-based deployment.
+
+Example:
+```bash
+dapr status -k
+
+  NAME                   NAMESPACE    HEALTHY  STATUS   REPLICAS  VERSION   AGE  CREATED
+  dapr-sentry            dapr-system  True     Running  1         1.7.0     17d  2022-03-15 09:29.45
+  dapr-dashboard         dapr-system  True     Running  1         0.9.0     17d  2022-03-15 09:29.45
+  dapr-sidecar-injector  dapr-system  True     Running  1         1.7.0     17d  2022-03-15 09:29.45
+  dapr-operator          dapr-system  True     Running  1         1.7.0     17d  2022-03-15 09:29.45
+  dapr-placement-server  dapr-system  True     Running  1         1.7.0     17d  2022-03-15 09:29.45
+⚠  Dapr root certificate of your Kubernetes cluster expires in 2 days. Expiry date: Mon, 04 Apr 2022 15:01:03 UTC.
+ Please see docs.dapr.io for certificate renewal instructions to avoid service interruptions.
+```
+
 ### Kubernetes video demo 
 Watch this video to show how to update mTLS certificates on Kubernetes
 


halspang:
> Can you please propose text that explains circuitbreakers and explain the relevant options?

I can give it a shot.

Circuit Breakers (CBs) are policies that are used when other applications/services/components are experiencing elevated failure rates. Their purpose is to monitor the requests and, when a certain criteria is met, shut off all traffic to the impacted service. This is to give the service time to recover from their outage instead of flooding them with events. The circuit breaker can also allow partial traffic through to see if the system has healed. Once successful requests start to occur, the CB can close and allow traffic to resume.

`MaxRequests`: The maximum number of requests allowed to pass through when the CB is half-open (recovering from failure). Defaults to 1.
`Interval`: The cyclical period of time used by the CB to clear its internal counts. If set to 0 seconds, this will never clear. Defaults to 0 seconds.
`Timeout`: The period of the open state (directly after failure) until the CB switches to half-open. Defaults to 60 seconds.
`Trip`: A Common Expression Language (CEL) statement that is evaluated by the CB. When the statement evaluates to true, the CB trips and becomes open. Default is `consecutiveFailures > 5`. 

> Sometimes the example uses general policies for retry and and circuitbreaking. Are these default policies that come for free with the runtime? Or must every policy be explicitly defined? If the latter the example needs to be updated (and the testdata in our repo is wrong).

Every policy has to be defined. In testdata, we define `general` as a timeout. So I don't think anything needs to be updated there. Though in the long run, it would be good for us to provide samples that people can use as a default policy. Note that they would still need to apply it themselves, I don't think we should force anything on them. They should just have an easy example to get started with.

```
timeouts:
      general: 5s
      important: 60s
      largeResponse: 10s
```

daprdocs/content/en/operations/resiliency/configure-policies.md
```diff
@@ -0,0 +1,282 @@
...
+linkTitle: "Resiliency Policies"
+weight: 4500
+description: "Configure Dapr error retries, timeouts and circuit breakers"
+---
+
+Resiliency is currently a preview feature. Before you can utilize resiliency policies you must first [enable the resiliency preview feature]({{<ref preview-features >}}).
+
+## Introduction
+
+- TODO: What is resiliency in Dapr?
```
greenie-msft: Maybe we could add something along the lines of:

Distributed applications are commonly comprised of many moving pieces - there could be dozens or even hundreds of instances for any given service. With this many moving pieces, the likelihood of a system failure increases. An instance can fail for any number of reasons, for example, hardware failures, overwhelming number of requests, application restarts/scale outs. Any of these events can cause a network call between services to fail. Having your application designed with the ability to detect and mitigate these failures allows for your application to  respond and recover quickly back to a functioning state.

daprdocs/content/en/operations/resiliency/configure-policies.md
```diff
@@ -0,0 +1,282 @@
...
+- TODO: What problems does it solve?
+
+## Overview
+
+A Dapr resiliency policy allow retry, timeout and circuitbreaker policies to be created and applied to particular targets including specific components and apps (service invocation calls to other apps).
```
greenie-msft:
```suggestion
A Dapr resiliency policy allows retry, timeout and circuitbreaker policies to be created and applied to particular targets including specific components and apps (service invocation calls to other apps).
```

daprdocs/content/en/operations/resiliency/configure-policies.md
```diff
@@ -0,0 +1,282 @@
...
+
+Additionally, resiliency policies can also be [scoped to specific apps]({{<ref "component-scopes.md#application-access-to-components-with-scopes">}}).
```
greenie-msft: @berndverst we need a section here to talk about service invocation specifically. The reason is that service invocation already has retries built-in. For every failed call, Dapr automatically retries up to 3 times: https://docs.dapr.io/developing-applications/building-blocks/service-invocation/service-invocation-overview/#retries

daprdocs/content/en/operations/resiliency/configure-policies.md
```diff
@@ -0,0 +1,282 @@
...
+spec:
+  policies:
+    # Timeouts are simple named durations.
+    timeouts:
+      general: 5s
+      important: 60s
+      largeResponse: 10s
+```
+
+#### Retries
```
greenie-msft: What do we mean here when we say "Defaults to X". If I decide to omit a value for a given  property, is it still applied implicitly with the default value?

daprdocs/content/en/operations/resiliency/configure-policies.md
```diff
@@ -0,0 +1,282 @@
...
+#### Apps
+
+Allows applying of `retry`, `timeout` and `circuitbreaker` policies to service invocation calls to other Dapr apps. Policy assignments are optional.
+
+Example
+```yaml
+specs:
+  targets:
+    apps:
+      appB:
```
greenie-msft: We need to mention that the value here is a Dapr App-ID and it's the policy applied to that service when it's the target ("callee")

daprdocs/content/en/operations/resiliency/configure-policies.md
```diff
@@ -0,0 +1,282 @@
...
+      myActorType:
+        timeout: general
+        retry: general
+        circuitBreaker: general
+        circuitBreakerScope: both
+        circuitBreakerCacheSize: 5000
+```
+#### Components
+
+Allows applying of `retry`, `timeout` and `circuitbreaker` policies to components operations. Policy assignments are optional. Policies can be applied for `outbound` operations (calls to the Dapr sidecar) or `inbound` (the sidecar calling your app). At this time, inbound only applies to PubSub and InputBinding components.
```
greenie-msft: "Policies can be applied for `outbound` operations (calls to the Dapr sidecar)"

This is actually sidecar calling the component, right? Not app calling the sidecar.

PR #2282, opened 2022-03-22
Remove dash in "dapr-init"

duglin:
I'm guessing it's a typo because the md file has the dash. But the command
itself doesn't


mukundansundar (approved):
lgtm

PR #2283, opened 2022-03-22
Minor updates to the golang pub-sub quickstart

duglin:
Minor updates to the golang pub-sub quickstart
- make the 'cd' command more clear by telling them to start from the clone
- add the sample output from the publisher
- say it's ok if the received events are out of order
- mention we're building the exe too


msfussell (approved):
LGTM

PR #2333, opened 2022-04-05
Fix link to dapr diagrams presentation

greenie-msft:

Fixes the button that links to the Dapr Diagrams Deck


msfussell (approved):
Thanks for fixing this

Issue #2354, opened 2022-04-10
Pub/Sub AWS SNS/SQS - document new concurrency mode metadata field


amimimor:
**What content needs to be created or modified?**

Pub/Sub AWS SNS/SQS - document new `concurrencyMode` metadata field usage

**Describe the solution you'd like**

concurrency mode should be added to the metadata fields description of the component

**Where should the new material be placed?**

**The associated pull request from dapr/dapr, dapr/components-contrib, or other Dapr code repos**

https://github.com/dapr/components-contrib/pull/1651

**Additional context**

msfussell:
Resolved

PR #2355, opened 2022-04-10
concurrency mode

amimimor:
Signed-off-by: Amit Mor <amitm@at-bay.com>

- [X] Commits are signed with Developer Certificate of Origin (DCO - [learn more](https://docs.dapr.io/contributing/contributing-overview/#developer-certificate-of-origin-signing-your-work))
- [X] [Read the contribution guide](https://docs.dapr.io/contributing/contributing-docs/)
- [X] Commands include options for Linux, MacOS, and Windows within codetabs
- [X] New file and folder names are globally unique
- [X] Page references use shortcodes instead of markdown or URL links
- [X] Images use HTML style and have alternative text
- [X] Places where multiple code/command options are given have codetabs

## Description

AWS Pub/Sub SNS/SQS component: add documentation about a new metadata field named `concurrencyMode` allowing the user to select either "parallel" or "single" publishing mode. As defined by this new field, upon receiving a bulk of message from SQS, the sidecar would publish the messages to the client with default mode of "parallel", i.e. concurrently, otherwise, publishing would be sequential, one message at a time.

## Issue reference

#2354

### DIFF
--- a/daprdocs/content/en/reference/components-reference/supported-pubsub/setup-aws-snssqs.md
+++ b/daprdocs/content/en/reference/components-reference/supported-pubsub/setup-aws-snssqs.md
@@ -82,7 +82,7 @@ The above example uses secrets as plain strings. It is recommended to use a secr
 | disableEntityManagement | N  | When set to true, SNS topics, SQS queues and the SQS subscriptions to SNS do not get created automatically. Default: `"false"` | `"true"`, `"false"`
 | disableDeleteOnRetryLimit | N  | When set to true, after retrying and failing of `messageRetryLimit` times processing a message, reset the message visibility timeout so that other consumers can try processing, instead of deleting the message from SQS (the default behvior). Default: `"false"` | `"true"`, `"false"`
 | assetsManagementTimeoutSeconds | N  | Amount of time in seconds, for an AWS asset management operation, before it times out and cancelled. Asset management operations are any operations performed on STS, SNS and SQS, except message publish and consume operations that implement the default Dapr component retry behavior. The value can be set to any non-negative float/integer. Default: `5` | `0.5`, `10`
-| concurrencyMode | N  | When a bulk of messages is pulled from SQS, call the subscriber sequencially (“single” message at a time), or concurrently (in “parallel”). Default: `"parallel"` | `"single"`, `"parallel"`
+| concurrencyMode | N  | When messages are received in bulk from SQS, call the subscriber sequentially (“single” message at a time), or concurrently (in “parallel”). Default: `"parallel"` | `"single"`, `"parallel"`
 
 
 * Dapr created SNS topic and SQS queue names conform with [AWS specifications](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/quotas-queues.html). By default, Dapr creates an SQS queue name based on the consumer `app-id`, therefore Dapr might perform name standardization to meet with AWS specifications.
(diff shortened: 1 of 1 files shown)


msfussell:
Per my comments in the code issue, is this just for AWS SQS? Also it would be good to provide a few more details like max number of messages (what happens if there are 22 messages for example) and can these each get sent to different message (topic) handlers. 

If this is general to several components, it may be worth highlighting this in the conceptual docs also.

amimimor:
Sure. I'll add those good points to the documentation. 
Per your question, the idea is based on the fact that the user can define the bulk size of messages that are pulled from SQS as a batch. Once that batch (or smaller) arrive, the component can now pass that bulk to the handler concurrently. In that sense, one SQS corresponds to one subscriber app. If I recall there is already such behavior in RabbitMQ.

daprdocs/content/en/reference/components-reference/supported-pubsub/setup-aws-snssqs.md
```diff
@@ -81,6 +82,7 @@ The above example uses secrets as plain strings. It is recommended to use a secr
 | disableEntityManagement | N  | When set to true, SNS topics, SQS queues and the SQS subscriptions to SNS do not get created automatically. Default: `"false"` | `"true"`, `"false"`
 | disableDeleteOnRetryLimit | N  | When set to true, after retrying and failing of `messageRetryLimit` times processing a message, reset the message visibility timeout so that other consumers can try processing, instead of deleting the message from SQS (the default behvior). Default: `"false"` | `"true"`, `"false"`
 | assetsManagementTimeoutSeconds | N  | Amount of time in seconds, for an AWS asset management operation, before it times out and cancelled. Asset management operations are any operations performed on STS, SNS and SQS, except message publish and consume operations that implement the default Dapr component retry behavior. The value can be set to any non-negative float/integer. Default: `5` | `0.5`, `10`
+| concurrencyMode | N  | When a bulk of messages is pulled from SQS, call the subscriber sequencially (“single” message at a time), or concurrently (in “parallel”). Default: `"parallel"` | `"single"`, `"parallel"`
```
berndverst:
```suggestion
| concurrencyMode | N  | When messages are received in bulk from SQS, call the subscriber sequentially (“single” message at a time), or concurrently (in “parallel”). Default: `"parallel"` | `"single"`, `"parallel"`
```

berndverst: Fix spelling / clarify this a bit.

berndverst (changes requested):
LGTM - just a small change

berndverst:
@msfussell I think what this does is... upon receiving a message batch it can process each message sequentially or in parallel -- this processing is entirely internal to the component, so it does not actually change anything about the size of the batch or the number of requests from SQS or from Dapr.

amimimor:
> LGTM - just a small change

Agreed. Also, we should somewhere document the change in behavior in v1.8.0 switching from sequential to parallel as default (I'm guessing in the release notes, but don't know how)

amimimor:
@berndverst @msfussell can you please rerun the checks as they fail on `validate-links` (while the "bad" link is ok)? thanks

msfussell:
> > LGTM - just a small change
> 
> Agreed. Also, we should somewhere document the change in behavior in v1.8.0 switching from sequential to parallel as default (I'm guessing in the release notes, but don't know how)

@amimimor - I suggest that you include this into the table as a note, so that this is clear on the change and then we can merge. I would suggest mentioning that from v1.8 the default has changed.

amimimor:
> > > LGTM - just a small change
> > 
> > 
> > Agreed. Also, we should somewhere document the change in behavior in v1.8.0 switching from sequential to parallel as default (I'm guessing in the release notes, but don't know how)
> 
> @amimimor - I suggest that you include this into the table as a note, so that this is clear on the change and then we can merge. I would suggest mentioning that from v1.8 the default has changed.

I've added it to the notes below the table. hope that makes sense. LMK if otherwise

msfussell (approved):
LGTM

Issue #2362, opened 2022-04-13
Quickstart Service Invocation tab for Javascript points at incorrect port creating exception


c835722:
**URL of the docs page**

https://docs.dapr.io/getting-started/quickstarts/serviceinvocation-quickstart/
Javascript tab

**How is it currently worded?**

dapr run --app-port 6001 --app-id order-processor --app-protocol http --dapr-http-port 3501 -- npm start

**How should it be worded?**

dapr run --app-port 5001 --app-id order-processor --app-protocol http --dapr-http-port 3501 -- npm start

**Screenshots**

**Additional context**

PR #2369, opened 2022-04-14
[quickstarts] fix port typo in pubsub and service invo

hhunter-ms:
## Description

* fix port typo for js in service invo
* fix port typo for python in pub/sub

## Issue reference

#2362


daprdocs/content/en/getting-started/quickstarts/pubsub-quickstart.md
```diff
@@ -702,7 +702,7 @@ go build app.go
 Run the `order-processor` subscriber service alongside a Dapr sidecar.
 
 ```bash
-dapr run --app-port 6002 --app-id order-processor --app-protocol http --dapr-http-port 3501 --components-path ../../../components -- go run app.go
+dapr run --app-port 6001 --app-id order-processor --app-protocol http --dapr-http-port 3501 --components-path ../../../components -- go run app.go
```
msfussell, line 705: The problem here is with the code. Validate for me, but the Go HTTP sample uses 6002 and the SDK sample uses 6001. They should both be the same otherwise fixing this docs issue with break the other one. so this is a code bug.

daprdocs/content/en/getting-started/quickstarts/serviceinvocation-quickstart.md
```diff
@@ -178,7 +178,7 @@ npm install
 Run the `order-processor` service alongside a Dapr sidecar.
 
 ```bash
-dapr run --app-port 6001 --app-id order-processor --app-protocol http --dapr-http-port 3501 -- npm start
+dapr run --app-port 5001 --app-id order-processor --app-protocol http --dapr-http-port 3501 -- npm start
```
msfussell, line 181: This one is correct and should be 5001. Why these are all so different is strange!

daprdocs/content/en/getting-started/quickstarts/pubsub-quickstart.md
hhunter-ms: since the quickstart focuses on the SDK, ok to update this to 6001? ...should we mention somewhere to follow the HTTP READMEs for the correct commands? at least until we make the code more consistent?

msfussell: We should fix the code to be the same.

greenie-msft: This still seems to be broken, is there an issue captured for QS to change these ports to be consistent?

msfussell: > This still seems to be broken, is there an issue captured for QS to change these ports to be consistent?

I put this issue into this issue https://github.com/dapr/quickstarts/issues/656 and marked this as P0.

daprdocs/content/en/getting-started/quickstarts/serviceinvocation-quickstart.md
akhilac1: If the port is being passed as --app-port wouldn't it be more appropriate to read it from the APP_PORT environment variable in the code vs. fixing the documentation?

msfussell: > If the port is being passed as --app-port wouldn't it be more appropriate to read it from the APP_PORT environment variable in the code vs. fixing the documentation?

Yes. Agreed on using APP_PORT. However the APP_PORT env variable is set by the Dapr CLI command and it would be best to make this consistent in the docs here.

msfussell (approved):
LGTM

PR #2397, opened 2022-04-26
Fix some minor config docs issues

duglin:
The config sample yaml used the wrong namespace and while in there I decided to
mention the name of the config instead of just its namespace since the two
are so similar, I didn't want people to think one was a typo.

And the monitoring sample had a typo w.r.t. enable/disable.

The text before the metrics sample talked about disabling it, but then the example shows it being enabled - felt a bit odd. So I add a line before the same to say that the sample is "enabling" it so to not confuse people.

Adding a missing "i" on "mult-write" in the state store docs


msfussell (approved):
LGTM

PR #2416, opened 2022-05-04
Fixes in howto-bindings

ItalyPaleAle:
- JS and curl samples were sending data as objects but sample apps expected ints
- Improved JS sample to correctly handle async code and make it more readable
- Fixed indentation in .NET code
- JavaScript has uppercased S


msfussell (approved):
LGTM.

PR #2423, opened 2022-05-05
Add missing upgrade step for 1.7

yaron2:
Without this extra step, an upgrade to 1.7 is incomplete and users can't apply Resiliency policies.


msfussell (approved):
LGTM

PR #2459, opened 2022-05-26
[AWS components] Add note for AWS K8s components

### LINKED ISSUE #2438: AWS Credentials Usage
All components that utilize AWS services need to be added with a clear note about the correct usage of AWS credentials when defining a component spec for the following scenarios:
1. If any IAM policy is attached the the running daprd container (standalone or as a container Pod), one *must not* provide AccessKey and SecretKeys, and AccessToken
2. If *not using* any attached IAM policies, use AccessKey, SecretKey & AccessToken

any document related to AWS components

**Additional context**
It has been observed several times, by the community and by this author, while using the sns/sqs AWS pubsub component, but this is true with any AWS components, that the AWS Go SDK fails to authorize access when the user passes in accessKey and secretKey in situations where the daprd process had already been granted an IAM policy through the various mechanisms of policy attachments. This results in calls to AWS assets to fail on authorization and crash the boot process of the component. 

The AWS library behavior could have been more verbose to its users but it doesn't so it might be tricky for users as they naively don't see a reason why their provided, working access and secret keys fail to allow them using the Dapr AWS components in such cases

hhunter-ms:
Signed-off-by: Hannah Hunter <hannahhunter@microsoft.com>

## Description

Add suggested note to AWS components - may need guidance on placement in articles.
Also edited capitalization inconsistency in an article.

## Issue reference

PR will close: #2438


daprdocs/content/en/operations/hosting/kubernetes/kubernetes-hybrid-clusters.md
```diff
@@ -101,7 +101,7 @@ In order to launch a Dapr application on Windows, you'll first need to create a
    ```
 
 ### Linux applications
-If you have already got a dapr application with runs on Linux, you'll still need to add affinity rules as above, but choose linux affinity instead.
+If you have already got a dapr application with runs on Linux, you'll still need to add affinity rules as above, but choose Linux affinity instead.
```
msfussell:
```suggestion
If you already a Dapr application with runs on Linux, you'll still need to add affinity rules as above, but choose Linux affinity instead.
```

msfussell (approved):
Another wonderful update!!

PR #2480, opened 2022-06-01
Match linkTitle case for HTTP

AaronCrawfis:
## Description

Consistency of HTTP and Http in link titles was incorrect. Fixing it to be HTTP


msfussell (approved):
LGTM

PR #2499, opened 2022-06-07
Add partial shortcode and partials for components along with data files

johnewart:
Signed-off-by: John Ewart <johnewart@microsoft.com>

## Description

Uses Hugo data files to generate component state / version tables, this will make it easier / possible to query things about the state of components. We could also potentially use this to keep a list of SDK features available, state, etc. 

Also helps to keep some of the feature flags straight (i.e actor support is now a logical 'and' between etag and transactions to avoid needing to update three columns). 

Basic change is that, for each component, there are YAML files in a `data/components/<component_type>/<component_group>.yaml` file, each type of component has a partial that describes the table layout and the fields being used along with a list of mappings between sub-title and group, and then a common `description` partial that gets used to replace the generic "table description" header that exists on all sub-listings. 

## Issue reference

No issue, just a proposed improvement to how we generate our docs so that the data can be used for other purposes; in this case I would like to be able to generate a list of components based on their state and then do some research using that data and make this reusable for others.


msfussell:
This looks good. The one difference currently is that on the existing pages it also builds a heading index on the right of the page.  See RHS of the images below. This is not a big deal, but it would be good to address for consistency in the docs. 

Existing page
image

New data drive page
image


Also there are new build failures below that will need to be looked at.

artursouza (approved):

msfussell:
@johnewart - We should get this in soon, since the component status is changing. For example there is a PR that was merged for the PostgresSQL now being Stable

daprdocs/layouts/partials/components/state-stores.html
```diff
@@ -0,0 +1,41 @@
...
+        <th>Since runtime version</th>
+    </tr>
+    {{ range sort $components "component" }}
+    <tr>
+        <td><a href="/reference/components-reference/supported-state-stores/{{ .link }}/" }}>{{ .component
+                }}</a>
+        </td>
+        <td>{{ if .features.crud }}✅{{ end }}</td>
+        <td>{{ if .features.transactions }}✅{{ end }}</td>
+        <td>{{ if .features.etag }}✅{{ end }}</td>
```
msfussell: Looks like you are missing code here for TTL and Actors columns. Am I correct here?

daprdocs/data/components/state_stores/aws.yaml
```diff
@@ -0,0 +1,11 @@
...
+  link: setup-dynamodb
+  state: Alpha
+  version: v1
+  since: "1.0"
+  features:
+    crud: true
+    transactions: false
+    etag: false
+    ttl: true
+    query: false
```
msfussell, line 11: You need and entry here for actor support for each state store.

msfussell:
@johnewart  - Found a couple of issues in the code here to address. PTAL.

daprdocs/layouts/partials/components/state-stores.html
johnewart: The actors column is below, the logical and of etags and transactions (which is what signals actor support, correct?) and yes, looks like I somehow missed out on the TTL column, I will fix that

greenie-msft (approved):

PR #2507, opened 2022-06-10
Clarify how Component configuration works

### LINKED ISSUE #2091: Redis Output Binding: metadata format documentation missing
## In what area(s)?

/area docs
/area bindings

## What version of Dapr?

1.5.1

## Expected Behavior

When invoking an output binding configured to use Redis I am unsure how to supply additional metadata about the item I am sending to the binding. There are no samples / documentation showing how to supply the metadata. Consequently I tried the following and expect the data to be handled.

```
POST http://127.0.0.1:5380/v1.0/bindings/receipts.binding
Content-Type: application/json

{
    "data": {
        "storeId": "12234",
        "customerName": "Simon W",
        "loyaltyId": "42",
        "orderItems": [
            {
                "menuItemId": 1,
                "quantity": 1
            },
            {
                "menuItemId": 2,
                "quantity": 1
            },
            {
                "menuItemId": 3,
                "quantity": 3
            }
        ]
    },
    "metadata": {
        "fileName" : "12234.json"
    },
    "operation": "create"
}
```

## Actual Behavior

When I submit this call I receive back:

```
{
  "errorCode": "ERR_INVOKE_OUTPUT_BINDING",
  "message": "error when invoke output binding test.binding: redis binding: missing key on write request metadata"
}
```

As there is no documentation or samples for this binding I have no idea how to achieve the outcome I expect. After digging around I stumbled across https://github.com/dapr/dapr/issues/772 but the links from that are broken. In a pair programming session I randomly tried `key` instead of `fileName` in the `metadata` block and the request succeeded.

## Steps to Reproduce the Problem

See above.
 
Additional documentation around bindings and their message formats would probably help avoid these types of issues.

## Release Note

RELEASE NOTE:

sjwaight:
The challenge for new comers to Dapr is understanding how Component configuration can change across Components of the same general type (i.e. state stores). RTFM of the Docs site and trial & error will mean most people will eventually grok this concept, however it would be great to call it out early in the Docs so people are aware they need to pay attention and that configuration AND API calls will likely change and not simply be transparent between Components.

## Issue reference

This PR will close: #2091. 

While this PR is not specifically for the Redis binding, it addresses what is actually a much more generic "gotcha" that people should be aware of between all Component types.

### DIFF
--- a/daprdocs/content/en/concepts/components-concept.md
+++ b/daprdocs/content/en/concepts/components-concept.md
@@ -12,6 +12,14 @@ Dapr uses a modular design where functionality is delivered as a component. Each
 
  You can get a list of current components available in the hosting environment using the `dapr components` CLI command.
 
+## Component configuration
+
+Components are configured at design-time via use of a CustomResourceDefinition (CRD) YAML file which is stored in either a `components/local` folder within your solution, or globally in the `.dapr` folder created when invoking `dapr init`.
+
+It is important to understand that configuration values, particularly spec metdata, can change between Components of the same general Component type (i.e. state stores), and that some design-time configuraion values can be overridden at runtime when making requests to the Component's API.
+
+If you are working with one Component type during development (i.e. Redis) and intend to switch to another on release (i.e. MySQL) you should review the differences between their configurations and API formats and ensure you understand any impact. Additionally, review the Component's spec, paying particular attention to sample payloads for requests.
+
 The following are the component types provided by Dapr:
 
 ## State stores
--- a/daprdocs/content/en/concepts/components-concept.md
+++ b/daprdocs/content/en/concepts/components-concept.md
@@ -12,7 +12,7 @@ Dapr uses a modular design where functionality is delivered as a component. Each
 
  You can get a list of current components available in the hosting environment using the `dapr components` CLI command.
 
-## Component configuration
+## Component specification
 
 Each component has a specification (or spec) that it conforms to. Components are configured at design-time with a YAML file which is stored in either a `components/local` folder within your solution, or globally in the `.dapr` folder created when invoking `dapr init`. These YAML files adhere to the generic [Dapr component schema]({{<ref "component-schema.md">}}), but each is specific to the component specification. 
 
--- a/daprdocs/content/en/concepts/components-concept.md
+++ b/daprdocs/content/en/concepts/components-concept.md
@@ -14,7 +14,7 @@ Dapr uses a modular design where functionality is delivered as a component. Each
 
 ## Component configuration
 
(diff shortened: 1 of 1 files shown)


daprdocs/content/en/concepts/components-concept.md
```diff
@@ -12,51 +12,59 @@ Dapr uses a modular design where functionality is delivered as a component. Each
 
  You can get a list of current components available in the hosting environment using the `dapr components` CLI command.
 
+## Component configuration
```
msfussell:
```suggestion
## Component specification
```

daprdocs/content/en/concepts/components-concept.md
```diff
@@ -12,51 +12,59 @@ Dapr uses a modular design where functionality is delivered as a component. Each
...
+
+Components are configured at design-time via use of a CustomResourceDefinition (CRD) YAML file which is stored in either a `components/local` folder within your solution, or globally in the `.dapr` folder created when invoking `dapr init`. These YAML files adhere to the generic [Dapr component schema]({{<ref "component-schema.md">}}).
```
msfussell:
```suggestion
Each component has a specification (or spec) that it conforms to. Components are configured at design-time with a YAML file which is stored in either a `components/local` folder within your solution, or globally in the `.dapr` folder created when invoking `dapr init`. These YAML files adhere to the generic [Dapr component schema]({{<ref "component-schema.md">}}), but each is specific to the component specification. 
```

daprdocs/content/en/concepts/components-concept.md
```diff
@@ -12,51 +12,59 @@ Dapr uses a modular design where functionality is delivered as a component. Each
...
+
+It is important to understand that configuration values, particularly spec metadata, can change between Components of the same general Component type (i.e. state stores), and that some design-time configuraion values can be overridden at runtime when making requests to a Component's API. As a result, it is strongly recommended to review [component specs]({{<ref "components-reference">}}), paying particular attention to sample payloads for requests used to interact with the component.
```
msfussell:
```suggestion
It is important to understand that the component spec values, particularly the spec `metadata`, can change between components of the same component type, for example between different state stores, and that some design-time spec values can be overridden at runtime when making requests to a component's API. As a result, it is strongly recommended to review a [component's specs]({{<ref "components-reference">}}), paying particular attention to the sample payloads for requests to set the metadata used to interact with the component.
```

msfussell:
@sjwaight - Thanks for this contributions. Really appreciate this clarity being added. I change configuration to spec values, because is already overloaded in Dapr, and hence we prefer to refer to component specs to avoid this confusion. Please review.

msfussell (approved):
Thanks for this contribution. Appreciate this. For future reference, when if you do another PR it need to have the sign-off from your, according to our public Apache v2.0 licence. You will see this as the DCO requirement. in effect this means adding -s <yourname> to the commit statement!

Issue #2433, opened 2022-05-11
Add new metadata for MQTT binding


yaron2:
Based on https://github.com/dapr/components-contrib/pull/1674, add the new topic metadata for the MQTT binding.

cc @lotuc

msfussell:
resolved

PR #2525, opened 2022-06-14
Add mqtt metadata spec and fix sendgrid json sample

yaron2:
Closes https://github.com/dapr/docs/issues/2433.

### DIFF
--- a/daprdocs/content/en/reference/components-reference/supported-bindings/mqtt.md
+++ b/daprdocs/content/en/reference/components-reference/supported-bindings/mqtt.md
@@ -138,7 +138,7 @@ This component supports **output binding** with the following operations:
 
 ## Set topic per-request
 
-You can specify the topic to send the message to on a per-request basis using request metadata:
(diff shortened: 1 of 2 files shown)


daprdocs/content/en/reference/components-reference/supported-bindings/mqtt.md
```diff
@@ -136,6 +136,20 @@ This component supports **output binding** with the following operations:
 
 - `create`
 
+## Set topic per-request
+
+You can specify the topic to send the message to on a per-request basis using request metadata:
```
msfussell: Q: How does this relate to the `topic` is the spec metadata. Does this override that topic for each request and what happens if I set both and they are different? Would be good to clarify this here.

yaron2: It overrides it.

msfussell (approved):
LGTM

Issue #2189, opened 2022-02-08
Provide guidance that a Cosmos DB in mongo mode cannot use the Dapr CosmosDB component


greenie-msft:
There needs to be guidance that a CosmosDB that uses the Mongo API cannot use the CosmosDB state store component.  Instead, developers  can use the MongoDB state store component with the CosmosDB that uses the Mongo API.

Notes should be added on these pages: 
- https://docs.dapr.io/reference/components-reference/supported-state-stores/setup-azure-cosmosdb/

PR #2544, opened 2022-06-20
Guide users to generic mongo db component for azure cosmos db with mo…

greenie-msft:
Added a sentence that instructs users using Azure Cosmos with Mongo API to use the generic Mongo component for a state store

## Issue reference
https://github.com/dapr/docs/issues/2189


berndverst:
Let's rename the component title to Azure Cosmos DB (SQL API). Every user of Cosmos DB knows they have to use other integrations to use anything that isn't SQL API.

greenie-msft:
> Let's rename the component title to Azure Cosmos DB (SQL API). Every user of Cosmos DB knows they have to use other integrations to use anything that isn't SQL API.

I've updated the Page Title, Link Title and Description to include (SQL API). Did the same for Bindings.

msfussell (approved):
LGTM

PR #2576, opened 2022-06-28
add docs for dapr annotate command

mukundansundar:
Signed-off-by: Mukundan Sundararajan <65565396+mukundansundar@users.noreply.github.com>

## Description

This PR is the docs PR for CLI command `dapr annotate`. 

Corresponding closed issue in CLI https://github.com/dapr/cli/issues/870. 
Corresponding PR: https://github.com/dapr/cli/pull/873


mukundansundar:
@jjcollinge I am seeing that the defaults are set to `-1`, but based on the code generally the defaults are ignored if it is `-1` for the `int` fields. So in that fashion, should we remove the default values from the default column for those `int` fields in the `annotate` command as this leads to some discrepancy between the annotation default values posted [here](https://v1-8.docs.dapr.io/reference/arguments-annotations-overview/)? 
Example: `dapr.io/sidecar-liveness-probe-delay-seconds` defaults to `3` based on the above linked page, but the value is said to default to `-1` in CLI which is actually discarded and not applied to the configuration. 

@msfussell Wanted your opinion on the slight difference in the annotation names, there are flags like `--liveness-probe-delay` which corresponds to `dapr.io/sidecar-liveness-probe-delay-seconds`. For convenience sake, the `sidecar` and `seconds` prefix and suffix are omitted as otherwise the flag will be extremely long. Similarly there are other sidecar based annotations that have the `sidecar` prefix omitted. IMO that is ok. What is your opinion? Should the docs be modified in anyway to tie the flag to the corresponding annotation to make it more clear? 

@yaron2 @jjcollinge @msfussell Have put this command as a preview feature command for now is that ok?

daprdocs/content/en/reference/cli/dapr-annotate.md
```diff
@@ -0,0 +1,84 @@
...
+| `--readiness-probe-delay` | | `-1` | The delay to use for the readiness probe in the sidecar. Read more [here](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes).|
+| `--readiness-probe-period` | | `-1` | The period to use for the readiness probe in the sidecar. Read more [here](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes). |
+| `--readiness-probe-threshold` | | `-1` | The threshold to use for the readiness probe in the sidecar. Read more [here](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes). |
+| `--readiness-probe-timeout` | | `-1` | The timeout to use for the readiness probe in the sidecar. Read more [here](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes). |
+| `--resource, -r` | | | The resource to target to annotate |
+
+{{% alert title="Warning" color="warning" %}}
+If an application ID is not provided using `--app-id, -a`, we will generate one using the format `<namespace>-<kind>-<name>`.
```
msfussell: I would have thought app-id is mandatory. And if not

```suggestion
If an application ID is not provided using `--app-id, -a`, and ID is generated using the format `<namespace>-<kind>-<name>`.
```

msfussell:
> @jjcollinge I am seeing that the defaults are set to `-1`, but based on the code generally the defaults are ignored if it is `-1` for the `int` fields. So in that fashion, should we remove the default values from the default column for those `int` fields in the `annotate` command as this leads to some discrepancy between the annotation default values posted [here](https://v1-8.docs.dapr.io/reference/arguments-annotations-overview/)? Example: `dapr.io/sidecar-liveness-probe-delay-seconds` defaults to `3` based on the above linked page, but the value is said to default to `-1` in CLI which is actually discarded and not applied to the configuration.
> 
> @msfussell Wanted your opinion on the slight difference in the annotation names, there are flags like `--liveness-probe-delay` which corresponds to `dapr.io/sidecar-liveness-probe-delay-seconds`. For convenience sake, the `sidecar` and `seconds` prefix and suffix are omitted as otherwise the flag will be extremely long. Similarly there are other sidecar based annotations that have the `sidecar` prefix omitted. IMO that is ok. What is your opinion? Should the docs be modified in anyway to tie the flag to the corresponding annotation to make it more clear?
> 
> @yaron2 @jjcollinge @msfussell Have put this command as a preview feature command for now is that ok?

@mukundansundar  - I would make the command and the annotations the same, then this is really specific. It is not as if we expect people to use this frequently. Also the odd this command is that this is now ONLY K8s specific and we really should avoid this, by having the -k option. This now is the only K8s command in the CLI. then annotations are a K8s thing,  however I would still prefer to "force" the -k option. WDYT? Finally if this is preview feature then you need to add this here, even for the CLI https://docs.dapr.io/operations/support/support-preview-features/

jjcollinge:
@mukundansundar thanks for raising that point. It definitely makes sense to try and keep these consistent - although I am not sure Cobra actually allows us to remove the defaults from the flags (or set a nil default) which makes this a bit tricky. If you know of a way then please let me know. If not then we have to provide a `int` default value - we could set it to the same defaults as the runtime if it has a default but this would mean that we would always have to set those annotations (or check if the value is the default value and then ignored it). Wdyt?
I'd be happy to add the `-k` flag and fail if not provided as @msfussell suggested.

daprdocs/content/en/reference/cli/dapr-annotate.md
jjcollinge: `app-id` is mandatory in the manifest but not in the CLI command. This is because you may be modifying an existing manifest that already has an app-id annotation - for instance `dapr annotate --log-level debug https://raw.githubusercontent.com/dapr/quickstarts/master/tutorials/hello-kubernetes/deploy/node.yaml` - this is just changing the log level of an existing dapr app so doesn't need to redefine `--app-id=nodeapp`. Rather than having to overwrite the app-id every time I just made it optional with a default convention if not provided by the CLI and not already present in the manifest. If you'd rather we didn't use this convention then I can make `app-id` flag required in the CLI. Let me know.

jjcollinge:
I've opened [a PR](https://github.com/dapr/cli/pull/1016) that adds the `-k` flag to the `annotate` command - it only supports Kubernetes so will fail if the `-k` flag is not present.

daprdocs/content/en/reference/cli/dapr-annotate.md
```diff
@@ -0,0 +1,84 @@
...
+| `--namespace, -n` | | | The namespace the resource target is in (can only be set if `--resource` is also set) |
+| `--readiness-probe-delay` | | `-1` | The delay to use for the readiness probe in the sidecar. Read more [here](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes).|
+| `--readiness-probe-period` | | `-1` | The period to use for the readiness probe in the sidecar. Read more [here](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes). |
+| `--readiness-probe-threshold` | | `-1` | The threshold to use for the readiness probe in the sidecar. Read more [here](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes). |
+| `--readiness-probe-timeout` | | `-1` | The timeout to use for the readiness probe in the sidecar. Read more [here](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes). |
+| `--resource, -r` | | | The resource to target to annotate |
```
shubham1172: TODO: Add other annotations from https://github.com/dapr/cli/issues/1018

mukundansundar: done

mukundansundar:
@msfussell Changed docs please review ...

daprdocs/content/en/reference/cli/dapr-annotate.md
```diff
@@ -0,0 +1,88 @@
+---
+type: docs
+title: "annotate CLI command reference"
+linkTitle: "annotate"
+description: "Add dapr annotatations to a Kubernetes configuration"
+---
+
+### Description
+
+Add dapr annotations to a Kubernetes configuration.
```
msfussell: I presume arguments here reflects this list https://v1-8.docs.dapr.io/reference/arguments-annotations-overview/ 
If so has this list been checked since I could not find dapr.io/sidecar-image as a value.

msfussell: If the above is true (the list) then I suggest the following update. 

```suggestion
Add Dapr annotations to a Kubernetes configuration. This enables you to change the Dapr annotations on a deployment other than those specified in the deployment file. See [Kubernetes annotations]({{< ref arguments-annotations-overview >}}) for a full description of each annotation.
```

daprdocs/content/en/reference/cli/dapr-annotate.md
```diff
@@ -0,0 +1,88 @@
...
+| `--namespace, -n` | | | The namespace the resource target is in (can only be set if `--resource` is also set) |
+| `--readiness-probe-delay` | | `-1` | The delay to use for the readiness probe in the sidecar. Read more [here](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes).|
+| `--readiness-probe-period` | | `-1` | The period to use for the readiness probe in the sidecar. Read more [here](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes). |
+| `--readiness-probe-threshold` | | `-1` | The threshold to use for the readiness probe in the sidecar. Read more [here](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes). |
+| `--readiness-probe-timeout` | | `-1` | The timeout to use for the readiness probe in the sidecar. Read more [here](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes). |
+| `--resource, -r` | | | The resource to target to annotate |
```
msfussell: What is --resource? When is this used?

msfussell:
@mukundansundar  - Reviewed the docs.

daprdocs/content/en/reference/cli/dapr-annotate.md
mukundansundar: See examples with the `-r` flag ... It is the Kubernetes deployment resource that is the target for the annotate command ... 
```
# Annotate multiple deployments by name in a chain
kubectl get deploy -o yaml | dapr annotate -k -r nodeapp - | dapr annotate -k -r pythonapp - | kubectl apply -f -
```

More examples in the file ...

daprdocs/content/en/reference/cli/dapr-annotate.md
mukundansundar: @msfussell I believe that this command as it exists is not for all the annotations as it exists. I am also thinking that all annotations should not be supported via this command. I am planning on creating a proposal to change this command to follow a helm based `--set` based format for 1.9. I will make note of the annotations needed to allowed to changed there.

daprdocs/content/en/operations/support/support-preview-features.md
```diff
@@ -15,7 +15,7 @@ For CLI there is no explicit opt-in, just the version that this was first made a
 ## Current preview features
 | Feature | Description | Setting | Documentation | Version introduced |
 | ---------- |-------------|---------|---------------|-----------------|
-| **--image-registry** flag with Dapr CLI| In self hosted mode you can set this flag to specify any private registry to pull the container images required to install Dapr| N/A | [init CLI command reference]({{<ref "dapr-init.md#self-hosted-environment" >}}) | v1.7 |
+| **--image-registry** flag in Dapr CLI| In self hosted mode you can set this flag to specify any private registry to pull the container images required to install Dapr| N/A | [CLI init command reference]({{<ref "dapr-init.md#self-hosted-environment" >}}) | v1.7 |
```
mukundansundar, line 18: leaving this as such for now only with a slight change in phrasing ... will have a discussion with Yaron after 1.8 release on whether to have Preview features or Preview commands in CLI or not ... 
cc @msfussell @yaron2

msfussell (approved):
LGTM

PR #2596, opened 2022-07-01
Add distributed lock api reference

yaron2:
^ title


daprdocs/content/en/reference/api/distributed_lock_api.md
```diff
@@ -0,0 +1,150 @@
...
+400  | Malformed request
+500  | Request failed
+
+### HTTP Request Body
+
+The lock endpoint receives the following JSON payload:
+
+```json
+{
+    "resourceId": "",
```
msfussell, line 45: It would be good to have a short explanation for each of these fields, especially what resourceId and lockOwner are used for. Need to tie this into the docs for each lock having a name which I presume is the lockOwner. Can the resource Id be anything? What am I expected to put here?

daprdocs/content/en/reference/api/distributed_lock_api.md
```diff
@@ -0,0 +1,150 @@
...
+```
+POST http://localhost:<daprPort>/v1.0-alpha1/lock/<storename>
+```
+
+#### URL Parameters
+
+Parameter | Description
+--------- | -----------
+`daprPort` | The Dapr port
+`storename` | The `metadata.name` field component file. Refer to the [component schema] ({{< ref component-schema.md>}})
```
msfussell, line 24: Is store name correct here? Will locks always be implemented by a store? Or should this be called <lockcomponent>?

daprdocs/content/en/reference/api/distributed_lock_api.md
```diff
@@ -0,0 +1,150 @@
...
+      }'
+
+{
+    "success": "true"
+}
+```
+
+## Unlock
+
+This endpoint lets you unlock an existing lock.
```
msfussell:
```suggestion
This endpoint lets you unlock an existing lock based on the lock owner and resource Id
```

daprdocs/content/en/reference/api/distributed_lock_api.md
```diff
@@ -0,0 +1,150 @@
...
+type: docs
+title: "Distributed Lock API reference"
+linkTitle: "Distributed Lock API"
+description: "Detailed documentation on the distributed lock API"
+weight: 660
+---
+
+## Lock
+
+This endpoint lets you acquire a lock.
```
msfussell:
```suggestion
This endpoint lets you acquire a lock but supplying a named lock owner and a resource ID.
```

daprdocs/content/en/reference/api/distributed_lock_api.md
```diff
@@ -0,0 +1,150 @@
...
+{
+    "status": 0
+}
+```
+
+The `status` field contains the following response codes:
+
+Code | Description
+---- | -----------
+0  | Success
```
msfussell, line 138: Why does this need to return an int? Makes it hard to understand and I would have expected a descriptive string.

daprdocs/content/en/reference/api/distributed_lock_api.md
yaron2: You will always need a lock store. This shouldn't be called out here (and it's not called out anywhere on any other API reference either)

daprdocs/content/en/reference/api/distributed_lock_api.md
yaron2: Actually, it's very common in RESTful API design to return status codes (much like HTTP status codes) that map to a specific behavior that is otherwise too verbose to be put in the response.

daprdocs/content/en/reference/api/distributed_lock_api.md
yaron2: Changed.

daprdocs/content/en/reference/api/distributed_lock_api.md
yaron2: Added

msfussell (approved):
LGTM

PR #2651, opened 2022-07-14
Explain that `namespaceName` must be a FQDN to Service Bus

ItalyPaleAle:
See https://github.com/dapr/dapr/issues/4904


mukundansundar (approved):
Lgtm

Issue #2650, opened 2022-07-13
error ~\\.dapr\\bin\\daprd.exe does not exist


charris-msft:
**Describe the issue**
when I run the command `dapr run --app-id order-processor --components-path ../../../components/ -- python3 app.py` I get the error `exec: "C:\\Users\\charris\\.dapr\\bin\\daprd.exe": file does not exist`

**URL of the docs**
https://docs.dapr.io/getting-started/quickstarts/statemanagement-quickstart/

**Expected content**
I expect the `order-processor` service to start without error.

**Screenshots**
![image](https://user-images.githubusercontent.com/74415662/178836771-6228832b-a9e4-4f15-adc0-a86fb2adc19a.png)

**Additional context**
OS: Windows 11 
Shell: PowerShell

paulyuk:
Thank you for the report @charris-msft.

It looks like you need the Dapr CLI, which is the assumption of all our quickstarts and dev exp.  Please try this:
https://docs.dapr.io/getting-started/

charris-msft:
Thanks @paulyuk!
You're right, I did need to install the CLI! I missed that in the pre-requisites.

After installing and going through the init process, I ran the `dapr run` command again and hit another error:

`exec: "python3": executable file not found in %!P(MISSING)ATH%!(NOVERB)`

![image](https://user-images.githubusercontent.com/74415662/178840642-b61c95eb-0e68-4520-96c2-b59495b35304.png)


But if I run either of the following commands it works:
`dapr run --app-id order-processor --components-path ../../../components/ -- py app.py`
`dapr run --app-id order-processor --components-path ../../../components/ -- python app.py`

paulyuk:
Right this tutorial is right or wrong optimized for Linux (including WSL on Windows or MacOS) and for that reason it's `python3`.  This also matches python.org's beginning guidance.  It perplexes me that Windows distro defaults to python or py.  https://stackoverflow.com/questions/39910730/python3-is-not-recognized-as-an-internal-or-external-command-operable-program

msfussell:
This is correct for Windows. Closing

PR #2656, opened 2022-07-15
[quickstarts] add note about python vs python3

hhunter-ms:
Signed-off-by: Hannah Hunter <hannahhunter@microsoft.com>

## Description

Add note suggesting using `python` instead of `python3` if using windows

## Issue reference

#2650


nyemade-uversky (approved):
LGTM. This should work fine as long as they have python 3 installed (as the quickstart suggests).

mukundansundar (approved):

PR #2684, opened 2022-07-27
Upmerge docs 1.8 to 1.9

hhunter-ms:
Routine docs upmerge: v1.8 to v1.9.


greenie-msft (approved):

msfussell:
@hhunter-ms - I see in this file daprdocs/content/en/operations/hosting/kubernetes/kubernetes-overview.md there is a reference to a PNG file that has changed. Have you also upmerged the changes to the diagrams in the images folder, since this will break otherwise? I could not see this in the list, but maybe I missed this.

PR #2693, opened 2022-07-30
Change terminology stable->stable candidate for candidates with cert …

tanvigour:
…test

Signed-off-by: tanvigour <tanvi.gour@gmail.com>

- [ ] [Read the contribution guide](https://docs.dapr.io/contributing/contributing-docs/)

## Description

The change reflects change in terminology for components with certification tests to be made stable in 1.9 like Blob storage to "stable candidate" instead of "stable" since technically they are not stable yet

### DIFF
--- a/daprdocs/data/components/state_stores/azure.yaml
+++ b/daprdocs/data/components/state_stores/azure.yaml
@@ -1,6 +1,6 @@
 - component: Azure Blob Storage
   link: setup-azure-blobstorage
-  state: Stable
+  state: Stable Candidate
   version: v1
   since: "1.0"
   features:
--- a/daprdocs/data/components/state_stores/azure.yaml
+++ b/daprdocs/data/components/state_stores/azure.yaml
@@ -1,6 +1,6 @@
 - component: Azure Blob Storage
   link: setup-azure-blobstorage
-  state: Stable Candidate
+  state: Stable
   version: v1
   since: "1.0"
   features:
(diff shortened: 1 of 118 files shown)


daprdocs/data/components/state_stores/azure.yaml
```diff
@@ -1,6 +1,6 @@
 - component: Azure Blob Storage
   link: setup-azure-blobstorage
-  state: Stable
```
artursouza, line 3: LGTM. Can we also make the changes in 1.8 branch for the components that are stable candidates already?

artursouza (approved):

daprdocs/data/components/state_stores/azure.yaml
artursouza: If a component is already stable, it can stay that way. My question is more about how we can track the beta components that are stable candidates.

artursouza (changes requested):

daprdocs/data/components/state_stores/azure.yaml
berndverst: Stable candidate has no official meaning - it's a process detail where we are announcing that the current Beta is considered complete and will be made stable in the future. Either we keep the current Beta designation and simply announce that the current Beta is a stable candidate (essentially the Beta release is a RC), or we get rid of this idea of stable candidate again.

tanvigour: > If a component is already stable, it can stay that way. My question is more about how we can track the beta components that are stable candidates.

ok, so since blob storage certification tests were added in 1.8, we should mark it as a "stable candidate " in 1.8 and stable in 1.9 correct?

PR #2710, opened 2022-08-09
Mark State Store components as Beta that were missed in 1.6 and 1.7 :(

berndverst:
Mark state store components as Beta that were missed in 1.6 and 1.7 :(

- Cassandra State Store
- CockroachDB State Store


greenie-msft (approved):

daprdocs/data/components/state_stores/generic.yaml
```diff
@@ -11,7 +11,7 @@
     query: false
 - component: Apache Cassandra
   link: setup-cassandra
-  state: Alpha
+  state: Beta
   version: v1
   since: "1.0"
```
greenie-msft, line 16: @berndverst do you need to update the "since" fields too?

msfussell: I  understand  the "since" field is supposed to indicate when this component was introduced and does not reflect changes in the component state. If it is then then there may be others that we have not done this with, when they move to Beta. I do suggest that we update the "since" field.

berndverst:
Closed in favor of #2757

PR #2712, opened 2022-08-10
Fix pubsub quickstart component name to match QS source code

greenie-msft:

Update PubSub Python Quickstart to match component name in the quickstart repo


hhunter-ms (approved):
lgtm

Issue #2730, opened 2022-08-19
Doc update for Unsubscribe configuration API to return response body


pravinpushkar:
**What content needs to be created or modified?**
Content that needed modification  - https://docs.dapr.io/reference/api/configuration_api/#unsubscribe-configuration

**Describe the solution you'd like**
Unsubscribe configuration API now returns a response body instead of empty response with response status code.

**Where should the new material be placed?**
https://docs.dapr.io/reference/api/configuration_api/#unsubscribe-configuration

**The associated pull request from dapr/dapr, dapr/components-contrib, or other Dapr code repos**
https://github.com/dapr/dapr/pull/5020

PR #2731, opened 2022-08-19
modify unsubscribe config api respose

pravinpushkar:
Signed-off-by: Pravin Pushkar <ppushkar@microsoft.com>

## Issue reference
#2730


pravinpushkar:
ping @msfussell @greenie-msft

greenie-msft (approved):

hhunter-ms:
@pravinpushkar sorry this fell off the radar - is this still applicable for 1.9? if so, does it also apply to 1.10, 1.11, etc?

msfussell:
@pravinpushkar If this is still relevant we should only apply this to v1.11 and then the update (v1) configuration apis. Can you PTAL and decide if the current docs still need updating?

pravinpushkar:
@msfussell Seems these changes are already merged as part of some different PR. Only the example section is missing in current dapr docs. I will try to modify the PR. We can make it as part of 1.12 as it is only example that we will add.

msfussell:
@pravinpushkar - So you updated this for v1.9. At this point it may not be worth updating v1.9. The question is whether any of this should apply to v1.11, which I presume not. In other words are the v1.11 reference docs for config currently correct?

pravinpushkar:
> @pravinpushkar - So you updated this for v1.9. At this point it may not be worth updating v1.9. The question is whether any of this should apply to v1.11, which I presume not. In other words are the v1.11 reference docs for config currently correct?

@msfussell The current docs for v1.11 are correct and has up to date information. Lets keep this PR open till I make modification in this PR(only to add example section) and point it to v1.12 instead of earlier releases. 

Edit -  In fact, **lets close this**. I will open a different PR for creating an example.

msfussell (approved):
LGTM

PR #2744, opened 2022-08-23
Fix links to broken middleware components

greenie-msft:

Fixes broken links of middleware components

### DIFF
--- a/daprdocs/data/components/middleware/http.yaml
+++ b/daprdocs/data/components/middleware/http.yaml
@@ -4,7 +4,7 @@
   version: v1
   description: Restricts the maximum number of allowed HTTP requests per second
 - component: OAuth2 Authorization Grant flow
-  link: https://tools.ietf.org/html/rfc6749#section-4.1
+  link: https://www.rfc-editor.org/rfc/rfc6749#section-4.1
   state: Alpha
   version: v1
   description: Enables the [OAuth2 Authorization Grant flow](https://tools.ietf.org/html/rfc6749#section-4.1) on a Web API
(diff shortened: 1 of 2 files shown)


daprdocs/data/components/middleware/http.yaml
```diff
@@ -24,22 +24,22 @@
   version: v1
   description: Applies [Rego/OPA Policies](https://www.openpolicyagent.org/) to incoming Dapr HTTP requests
 - component: Sentinel
-  link: middleware-sentinel
+  link: /reference/components-reference/supported-middleware/middleware-sentinel
```
msfussell, line 27: Pretty sure that these is not supposed to be direct links to the docs. For example the https://docs.dapr.io/reference/components-reference/supported-middleware/middleware-sentinel/ link works currently sentinel and this is just "link: middleware-sentinel" rather than link: /reference/components-reference/supported-middleware/middleware-sentinel

daprdocs/layouts/partials/components/middleware.html
```diff
@@ -13,7 +13,7 @@ <h3>{{ $group }}</h3>
     </tr>
     {{ range sort $components "component" }}
     <tr>
-        <td><a href="/reference/components-reference/supported-middleware/{{ .link }}/" }}>{{ .component
+        <td><a href="{{ .link }}/" }}>{{ .component
```
msfussell, line 16: Changing this does require me to provide a full path above. Is that what we want, as opposed to just the end of the path?

msfussell (approved):

Issue #2750, opened 2022-08-25
fix typo


RcXu:
**URL of the docs page**
[](https://docs.dapr.io/reference/cli/dapr-mtls/dapr-mtls-renew-certificate/)
https://docs.dapr.io/reference/cli/dapr-mtls/dapr-mtls-renew-certificate/

**How is it currently worded?**
restarts the Dapr control place services.

**How should it be worded?**
restarts the Dapr control plane services.

**Screenshots**

<img width="922" alt="WX20220825-183509@2x" src="https://user-images.githubusercontent.com/17242622/186643147-98c99f67-e2c5-4041-8089-8d93e5f6d404.png">

PR #2751, opened 2022-08-25
fix typo

RcXu:
## Description

`Generates new root and issuer certificates for the Kubernetes cluster with a given validity time and restarts the Dapr control place services.`

there is a typo for the words "Dapr control plane",  instead "Dapr control place"

## Issue reference

#[[issue number](https://github.com/dapr/docs/issues/2750)]


msfussell (approved):
Thanks!

PR #2753, opened 2022-08-26
RethinkDB has no transactions, etags

berndverst:
Signed-off-by: Bernd Verst <4535280+berndverst@users.noreply.github.com>

RethinkDB has no transactions, etags


daixiang0:
Suggest that add version prefix to the title.

greenie-msft (approved):

PR #2755, opened 2022-08-26
RethinkDB now Beta, but has no transactions, etags

berndverst:
RethinkDB is now Beta

RethinkDB does not support transactions, etags

### DIFF
--- a/daprdocs/content/en/reference/components-reference/supported-state-stores/setup-rethinkdb.md
+++ b/daprdocs/content/en/reference/components-reference/supported-state-stores/setup-rethinkdb.md
@@ -38,16 +38,7 @@ spec:
 The above example uses secrets as plain strings. It is recommended to use a secret store for the secrets, as described [here]({{< ref component-secrets.md >}}).
 {{% /alert %}}
 
-If you wish to use RethinkDB as an actor store, append the following to the YAML.
-
-```yaml
-  - name: actorStateStore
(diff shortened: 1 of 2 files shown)


withinboredom:
> RethinkDB does not support transactions, etags

A single document write/update is atomic (if read-mode is `single` or `majority`) and multi-document updates are fairly straightforward to implement using 2-phase commit or [RAMP transactions](http://www.bailis.org/papers/ramp-sigmod2014.pdf).

Etags can be emulated with a last-modified key (set to r.now() on write/update) and a hash over it.

It has been an extremely long time since I've used rethinkdb daily. So my knowledge/memory may be wrong.

berndverst:
@withinboredom as the maintainer of components-contrib it is my obligation to ensure that current capabilities are accurately documented.

These capabilities are neither natively supported nor does the Dapr component currently support this.

Yes Dapr could fake native multi-document transaction support - something RethinkDB doesn't officially support.
The current Etag hashing implementation also does not pass the Etag conformance tests and officially RethinkDB has no Etag support or similar.

I will be open to PRs for someone to adequately implement ETag support or Transactions in the future. As it stands this component is not compliant with the Dapr ETag and Transaction API guarantees.

Feel free to open a PR or issue in Dapr/Contrib.

greenie-msft (approved):

PR #2756, opened 2022-08-29
Marks missing components as beta and stable in v1.7

berndverst:
This marks missing components as beta and stable in v1.7.

This is based on the new beta/stable component certification as outlined in my proposal which has been agreed upon.

Please merge after #2753


greenie-msft (approved):
LGTM

PR #2777, opened 2022-09-09
optimize: add component schema doc

1046102779:
/cc @msfussell


1046102779:
cc @yaron2 @artursouza

PR #2822, opened 2022-09-23
Updated docs for dapr/components-contrib#2104

ItalyPaleAle:
New metadata option for MQTT.

See dapr/components-contrib#2104


msfussell (approved):

msfussell:
@ItalyPaleAle - One question. Do you think any part of the conversation from the issue, should be covered in the usage section, or is there enough info on the metadata table description. Just making sure that the use of the new Producer field is clear (given the backward compatibility)

ItalyPaleAle:
@msfussell it seems to me there could be enough info, but WDYT? As someone who wasn't involved like I was, is this clear enough to you?

msfussell (approved):
LGTM

PR #2832, opened 2022-09-26
Add AWS S3 presign url

### LINKED ISSUE #2827: S3 Presign
S3 binding needs un update for the new presignTTL metadata.

yaron2:
Closes #2827

### DIFF
--- a/daprdocs/content/en/reference/components-reference/supported-bindings/s3.md
+++ b/daprdocs/content/en/reference/components-reference/supported-bindings/s3.md
@@ -256,6 +256,39 @@ The response body will contain the following JSON:
 }
 ```
 
+#### Presign an existing object
+
+To presign an existing S3 object with a specified time-to-live, use the `presignTTL` and `key` metadata keys on a `presign` request.
+Valid values for `presignTTL` are [Go duration strings](https://pkg.go.dev/maze.io/x/duration#:~:text=A%20duration%20string%20is%20a,w%22%2C%20%22y%22).
+
+{{< tabs Windows Linux >}}
+
+  {{% codetab %}}
+  ```bash
+  curl -d "{ \"operation\": \"presign\", \"metadata\": { \"presignTTL\": \"15m\", \"key\": \"my-test-file.txt\" } }" \
+        http://localhost:<dapr-port>/v1.0/bindings/<binding-name>
+  ```
+  {{% /codetab %}}
+
+  {{% codetab %}}
+  ```bash
+  curl -d '{ "operation": "presign", "metadata": { "presignTTL": "15m", "key": "my-test-file.txt" } }' \
+        http://localhost:<dapr-port>/v1.0/bindings/<binding-name>
+  ```
+  {{% /codetab %}}
+
+{{< /tabs >}}
+
+##### Response
+
+The response body contains the following example JSON:
+
+```json
+{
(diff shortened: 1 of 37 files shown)


daprdocs/content/en/reference/components-reference/supported-bindings/s3.md
```diff
@@ -103,6 +103,41 @@ To perform a create operation, invoke the AWS S3 binding with a `POST` method an
...
+  curl -d '{ "operation": "create", "data": "Hello World", "metadata": { "presignTTL": "15m" } }' \
+        http://localhost:<dapr-port>/v1.0/bindings/<binding-name>
+  ```
+  {{% /codetab %}}
+
+{{< /tabs >}}
+
+##### Response
+
+The response body will contain the following example JSON:
```
msfussell:
```suggestion
The response body contains the following example JSON:
```

daprdocs/content/en/reference/components-reference/supported-bindings/s3.md
```diff
@@ -103,6 +103,41 @@ To perform a create operation, invoke the AWS S3 binding with a `POST` method an
 }
 ```
 
+#### Share object with a presigned URL
+
+To presign an object with a specified time-to-live, use the `presignTTL` metadata key on a `create` request.
```
msfussell, line 105: Can you add the "presignTTL" metadata to the 
1) Component example https://docs.dapr.io/reference/components-reference/supported-bindings/s3/#component-format
2) The spec metadata fields table with a description/example https://docs.dapr.io/reference/components-reference/supported-bindings/s3/#spec-metadata-fields

yaron2: No, since this is a per-request metadata. This isn't supported on the YAML metadata.

msfussell: Ok for 1. but is still needed for 2. (the table) for discoverability and understanding. Same as for Filepath https://github.com/dapr/components-contrib/blob/ad5397ea25489943acf5156c8af764e6f79ea788/bindings/aws/s3/s3.go#L73

yaron2: That's a good catch, `filePath` should have been removed from both the Component YAML and the table as it's not a spec metadata. Pushed a commit to remove this.

msfussell: There is still a discoverability issue here that you can provide "filePath" and "PresignTTL" on the Create operation otherwise the description/usage info is lost e.g. "If set, create operations that contain empty data fields will attempt to upload a file from the supplied path" and not clear on usage. 

So if you do not think this should be in the metadata field then add this as a table under the Create operation along with "filePath"

yaron2: > So if you do not think this should be in the metadata field

Luckily, this is an absolute decision and no thinking is required here - this simply can't be on the table as it's not supported as a spec metadata. Had it been there, it would be false information.

I'm not sure I get the discoverability issue - there are dedicated sections and sub-headings in this doc for how to presign URLs and upload from a file path, with clear usage examples that show the keys used and the expected response. What else is needed?

msfussell: put this on line 228 & 229

```
##### Upload from file path
To upload a file from a supplied path (relative or absolute), use the `filepath` metadata key on a `create` request that contains empty `data` fields
```

yaron2: Done

msfussell (approved):
LGTM

PR #2880, opened 2022-10-10
Remove outdated comment on creating a middleware

ItalyPaleAle:
Removing one paragraph with outdated information. No other component type had a similar comment, so it made sense to remove it rather than updating it.

Up-to-date information can be found on the GitHub repo: https://github.com/dapr/components-contrib/blob/master/docs/developing-component.md

Also, users can now leverage pluggable components rather than shipping things in the runtime.


msfussell (approved):
LGTM

Issue #2738, opened 2022-08-20
Subscription Methods: Unclear/missing information


alecor191:
**Describe the issue**

The *Declarative Subscriptions* doc suggests that using the Dapr SDKs requires two steps to be able to subscribe to a Dapr pubsub:

1. Component file with `kind: Subscription`
2. Use SDK in the app to act when receiving messages from pubsub

However, I used both Python SDK as well as JS SDK to subscribe to pubsub and it worked fine without step 1. above (i.e. no Yaml file used).

I think it would be great if the following items could be clarified in the docs:

- Under what circumstances is the yaml file required? (e.g. I assume that I need it if I want to specify `scopes`?)
- How is the `route` value in the yaml file used when leveraging SDKs. Looking at the referenced doc, neither in JS nor in Python code samples, I see a reference to the `route: /checkout` route from the Yaml example:

  ```javascript
  //Subscribe to a topic
  await server.pubsub.subscribe("pubsub", "orders", async (orderId) => {
      console.log(`Subscriber received: ${JSON.stringify(orderId)}`)
  });
  await server.startServer();
  ```
  
  ```python
  #Subscribe to a topic 
  @app.subscribe(pubsub_name='pubsub', topic='orders')
  def mytopic(event: v1.Event) -> None:
      data = json.loads(event.Data())
      logging.info('Subscriber received: ' + str(data))
  
  app.run(6002)
  ```

  Further the doc mentions:

  > The `/checkout` endpoint matches the `route` defined in the subscriptions and this is where Dapr sends all topic messages to

  But I can't see the JS/Python apps reference anything related to `/checkout`
 

**URL of the docs**

https://docs.dapr.io/developing-applications/building-blocks/pubsub/subscription-methods/

**Expected content**

- Provide context under what circumstances the yaml file (`kind: Subscription`) is required.
- If SDK + Yaml file are used, please explain the role of `route` setting.

**Screenshots**

N/A.

msfussell:
@alecor191 - Can you review the attached PR and see if this fits with you expectation of the change needed?

alecor191:
Thanks @msfussell! I added my comments to the PR.

hhunter-ms:
This issue has been resolved - closing!

PR #2896, opened 2022-10-14
[subscriptions doc] update py, go, and js examples

hhunter-ms:
## Description

Update python, go, and javascript declarative code examples to match yaml example

## Issue reference

#2738


msfussell:
@hhunter-ms - These are good updates. Did you address the question in the issues from @alert01

hhunter-ms:
@msfussell I'm hoping these updates help answer those questions. Would love feedback @alecor191

alecor191:
Thanks a lot @hhunter-ms for looking into this! 

I have a few questiions about the changes:

- Python changes: should `route='route'` be `route='checkout'` instead, considering that the sample talks about the `/checkout` route?
- Python changes: What was the reason for changing `@app.subscribe(pubsub_name=...` to `@app.subscribe(pubsubname=...`? Isn't `pubsub_name` argument name correct?
- It would be great if we could add an explanation under what circumstances the yaml file (`kind: Subscription`) is required, as for me it worked also without it (see #2738 for details)

hhunter-ms:
> * Python changes: should `route='route'` be `route='checkout'` instead, considering that the sample talks about the `/checkout` route?

I believe so - I'll update that

> * Python changes: What was the reason for changing `@app.subscribe(pubsub_name=...` to `@app.subscribe(pubsubname=...`? Isn't `pubsub_name` argument name correct?

I changed this to match the [subscription example in the pub/sub quickstart for python](https://docs.dapr.io/getting-started/quickstarts/pubsub-quickstart/#step-3-subscribe-to-topics)... What do you think, @msfussell?

> * It would be great if we could add an explanation under what circumstances the yaml file (`kind: Subscription`) is required, as for me it worked also without it (see [Subscription Methods: Unclear/missing information #2738](https://github.com/dapr/docs/issues/2738) for details)

I'll work on this... Thanks for your feedback @alecor191!

hhunter-ms:
@alecor191 It looks like a lot of the confusion around the declarative subscriptions and why the YAML is important is because the code snippets offered as declarative examples were actually programmatic examples. I'm updating the declarative examples to actually reflect what the code would look like listening to the YAML. 

Hopefully that helps!

msfussell (approved):
LGTM

Issue #2861, opened 2022-10-05
Migrate to Google Analytics 4 (GA4)


chalin:
This issue is part of a CNCF-wide effort to upgrade project websites to GA4 since **Google has deprecated Universal Analytics** (UA). For more details, see:

- https://github.com/cncf/techdocs/issues/108, including reference to a suggested [migration process](https://github.com/cncf/techdocs/blob/main/docs/analytics.md#migrating-to-google-analytics-4), which we've used to [migrate a few CNCF projects](https://docs.google.com/spreadsheets/d/1Mx4LhdI2Un-rvGMI73SlHxQH9D2HABAJclMB3dd6lnA) already.

### Tasks

- [x] From the website data stream (`G-60C6Q1ETC1`),
    configure cross-domain measurement for domains _ending with_ **docs.dapr.io**
- [x] Connect `UA-58468480-3` from GA4 stream
- [ ] Configure GA4 website ID ([gtag.js]) for each release:
  - [x] v1.10 - #2945
  - [x] v1.9 - #2941
  - [x] v1.8 - #2950
  - [x] v1.7 - #2962
  - [x] v1.6 - #2963
  - [ ] ... (list to be completed) ...
- [x] Confirm that the GA4 site tag is receiving events -- [tagassistant.google.com](https://tagassistant.google.com/) was used to confirm
- [x] Confirm that the UA site tag, `UA-58468480-3`, is still receiving events

Notes:

- For the main repo issue, see https://github.com/dapr/website/issues/71
- Each doc version is under it's own branch, and each is separately configured to use the UA ID
- Related: #2943

/cc @caniszczyk @nate-double-u

---

Current website analytics info: `UA-149338238-3` registered via old `analytics.js` library.

[connect]: https://support.google.com/analytics/answer/9973999
[gtag.js]: https://developers.google.com/analytics/devguides/collection/gtagjs
[Netlify snippet injection]: https://docs.netlify.com/site-deploys/post-processing/snippet-injection/

chalin:
I understand that this is the repo for docs only. Still I'll be using this as a tracking issue for the migration to GA4 of the blog and main site too.

chalin:
Each doc version is under its own branch, and has its own individually configured UA ID. It'll be tedious to update all of those.

Do maintainers know if Azure Static Web Apps supports something like [Netlify snippet injection]? That would make the update to GA4 much easier for now, and in the future if the ID needs to be updated.

[...]

PR #2941, opened 2022-11-07
Use Google Analytics 4 ID for v1.9

chalin:
## Issue reference

- Contributes to #2861

## Description

This is the first PR in a series, as we migrate to the new Dapr GA4 site tag.

**IMPORTANT**:
> While we update the GA ID through the various doc versions/branches incrementally, I'll have to [connect] the GA4 ID to the original docs UA ID. Since the GA4 ID is _the_ ID for all Dapr things web, this means that **all web events**, including page_views etc. **from the main website, will be sent to the UA ID**.
>
> Please **confirm that this is ok with you** @berndverst et al.

/cc @berndverst @nate-double-u

[connect]: https://support.google.com/analytics/answer/9973999

### DIFF
--- a/daprdocs/config.toml
+++ b/daprdocs/config.toml
@@ -27,7 +27,7 @@ disableKinds = ["taxonomy", "term"]
 
 # Google Analytics
 [services.googleAnalytics]
-id = "UA-149338238-3"
+id = "G-60C6Q1ETC1"
 
 # Mounts
 [module]


chalin:
@berndverst - if you'd prefer setting the ID through a build environment variable (as discussed in the Dapr maintainers list thread), rather than explicitly as is done in this PR, let me know. In that case, I'll submit another PR.

Regardless, I'll still need for you to confirm that you're ok with the docs UA ID being connected from the GA4 web data-stream (see the opening comment for details).

/cc @msfussell

msfussell:
@chalin - We are good with "'ll still need for you to confirm that you're ok with the docs UA ID being connected from the GA4 web data-stream (see the opening comment for details)." as the docs maintainers. I suggest merging this PR for now.

msfussell (approved):
LGTM

PR #3031, opened 2022-12-27
Fixed issues with a11y on component list pages

ItalyPaleAle:
While working on something unrelated, I found some issues in the pages that listed components. Mainly related to accessibility (images and emojis used to convey important information but did not have the required tags for screen readers), but a few other fixes are in this PR too.


msfussell (approved):
LTGM. Thanks for taking the time to do this.

PR #3033, opened 2022-12-27
[1.10] Update docs for PostgreSQL state store

### LINKED ISSUE #2966: Update reference for state.postgres
The reference for the [Postgres state store](https://docs.dapr.io/reference/components-reference/supported-state-stores/setup-postgresql/) needs to be updated for 1.10 because of dapr/components-contrib#2302 and dapr/components-contrib#2379

In particular:

- Add note that TTLs are now supported (in an emulated way). Similar to how it's documented for the [Oracle DB state store](https://docs.dapr.io/reference/components-reference/supported-state-stores/setup-oracledatabase/)
- Document new metadata option `cleanupIntervalInSeconds` for the background "garbage collector" that removes expired records. The default value is 3600 seconds (1hr), and a value <= 0 means that no GC is performed (useful if the user knows they'll never use TTLs)
- Document new metadata option `timeoutInSeconds` for timeouts for network operations. The default value is 20 (seconds)
- Document missing metadata options, that were added in the past but not documented:
   - `tableName`, which defaults to `"state"` (can be in the format `table` or `schema.table`)
   - `connectionMaxIdleTime`, which causes connections to be closed automatically if idle for N seconds (default is 0, which means connections are not closed)

ItalyPaleAle:
Update the component's docs due to changes introduced in dapr/components-contrib#2302 and dapr/components-contrib#2379

Fixes #2966

### DIFF
--- a/daprdocs/content/en/reference/components-reference/supported-state-stores/setup-postgresql.md
+++ b/daprdocs/content/en/reference/components-reference/supported-state-stores/setup-postgresql.md
@@ -7,11 +7,13 @@ aliases:
   - "/operations/components/setup-state-store/supported-state-stores/setup-postgresql/"
 ---
 
+This component allows using PostgreSQL (Postgres) as state store for Dapr.
+
 ## Create a Dapr component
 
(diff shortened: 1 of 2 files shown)


msfussell (approved):
LGTM

PR #3059, opened 2023-01-13
[1.10] Rename mqtt components to mqtt3

ItalyPaleAle:
See https://github.com/dapr/components-contrib/issues/2359

The old component names are still available as aliases

### DIFF
--- a/daprdocs/content/en/operations/components/component-schema.md
+++ b/daprdocs/content/en/operations/components/component-schema.md
@@ -44,7 +44,7 @@ spec:
 
 ### Special metadata values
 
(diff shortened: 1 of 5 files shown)


hhunter-ms (approved):
lgtm

mukundansundar:
@ItalyPaleAle  
> The old component names are still available as aliases

I think this needs to be mentioned in the docs somewhere ....

ItalyPaleAle:
We normally don't do that when we update the name of a component, so not to encourage people to use the old name. When we've done this (and it's happened a few times already), we just updated the docs to use the new name.

msfussell:
> We normally don't do that when we update the name of a component, so not to encourage people to use the old name. When we've done this (and it's happened a few times already), we just updated the docs to use the new name.

This can be mentioned in the release notes. Will the old alias always continue to work, no intention of deprecating?

ItalyPaleAle:
> Will the old alias always continue to work, no intention of deprecating?

No intention, and personally I see no reason why we would.

PR #3100, opened 2023-01-30
Add documentation for using `automountServiceAccountToken: false`

### LINKED ISSUE #3008: Add documentation for using `automountServiceAccountToken: false`

Starting with Dapr 1.10, users can now run pods with Dapr where they can set `automountServiceAccountToken: false` in their Deployment spec.

By default, Kubernetes mounts a service account token in each container at the path `/var/run/secrets/kubernetes.io/serviceaccount/token` . This token has permission to invoke the Kubernetes APIs (with whatever permissions are assigned to the token, which are fairly broad for tokens used by Dapr), and for some users this may not be desirable as it _could_ increase the attack surface in some cases. This behavior, which is normally on by default, can be disabled in each Pod/Deployment by setting `automountServiceAccountToken: false` in the spec.

With Dapr 1.10, users are encourage to set `automountServiceAccountToken: false` in their Deployment YAML's unless their application needs a Kubernetes service account token for some reasons. Reasons include:

- Using a Kubernetes Dapr component, for example the Kubernetes event binding or the Kubernetes secret store
   - Note that using the Kubernetes secret store to store component initialization secrets is unaffected by this, as that's read by the Dapr Operator and not the Dapr sidecars. This impacts only users who are using components of type `secretstores.kubernetes` to retrieve secrets at runtime
- Their own app requiring interacting with the Kubernetes APIs.

Document the fact that users can optionally set `automountServiceAccountToken: false` in their Deployment YAML's, and when that's recommended, in the Dapr docs.

Possible target could be the Kubernetes page: https://docs.dapr.io/operations/hosting/kubernetes/kubernetes-production/

dapr/dapr#5371

ItalyPaleAle:
Fixes #3008

### DIFF
--- a/daprdocs/content/en/operations/hosting/kubernetes/kubernetes-production.md
+++ b/daprdocs/content/en/operations/hosting/kubernetes/kubernetes-production.md
@@ -196,6 +196,20 @@ It is recommended that a production-ready deployment includes the following sett
 
 6. Dapr also supports **scoping components for certain applications**. This is not a required practice, and can be enabled according to your security needs. See [here]({{< ref "component-scopes.md" >}}) for more info.
 
+## Service account tokens
+
+By default, Kubernetes mounts a volume containing a [Service Account token](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/) in each container. Applications can use this token, whose permissions vary depending on the configuration of the cluster and namespace, among other things, to perform API calls against the Kubernetes control plane.
+
+When creating a new Pod (or a Deployment, StatefulSet, Job, etc), you can disable auto-mounting the Service Account token by setting `automountServiceAccountToken: false` in your pod's spec.
+
+We recommend that you consider deploying your apps with `automountServiceAccountToken: false` to improve the security posture of your pods, unless your apps depend on having a Service Account token. For example, you may need a Service Account token if:
+
+- You are using Dapr components that interact with the Kubernetes APIs, for example the [Kubernetes secret store]({{< ref "kubernetes-secret-store.md" >}}) or the [Kubernetes Events binding]{{< ref "kubernetes-events.md" >}}).  
+  Note that initializing Dapr components using [component secrets]({{< ref "component-secrets.md" >}}) stored as Kubernetes secrets does **not** require a Service Account token, so you can still set `automountServiceAccountToken: false` in this case. Only calling the Kubernetes secret store at runtime, using the [Secrets management]({{< ref "secrets-overview.md" >}}) building block, is impacted.
+- Your own application needs to interact with the Kubernetes APIs.
+
+Because of the reasons above, Dapr does not set `automountServiceAccountToken: false` automatically for you. However, in all situations where the Service Account is not required by your solution, we do recommend that you set this option in the spec of your pods.
+
 ## Tracing and metrics configuration
 
 Dapr has tracing and metrics enabled by default. It is *recommended* that you set up distributed tracing and metrics for your applications and the Dapr control plane in production.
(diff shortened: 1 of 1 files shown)


daprdocs/content/en/operations/hosting/kubernetes/kubernetes-production.md
```diff
@@ -196,6 +196,20 @@ It is recommended that a production-ready deployment includes the following sett
 
 6. Dapr also supports **scoping components for certain applications**. This is not a required practice, and can be enabled according to your security needs. See [here]({{< ref "component-scopes.md" >}}) for more info.
 
+## Service account tokens
+
+By default, Kubernetes mounts a volume containing a [Service Account token](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/) in each container. Applications can use this token, whose permissions vary depending on the configuration of the cluster and namespace, among other things, to perform API calls against the Kubernetes control plane.
+
+When creating a new Pod (or a Deployment, StatefulSet, Job, etc), you can disable auto-mounting the Service Account token by setting `automountServiceAccountToken: false` in your pod's spec.
+
+We recommend that you consider deploying your apps with `automountServiceAccountToken: false` to improve the security posture of your pods, unless your apps depend on having a Service Account token. For example, you may need a Service Account token if:
```
msfussell:
```suggestion
It is recommended that you consider deploying your apps with `automountServiceAccountToken: false` to improve the security posture of your pods, unless your apps depend on having a Service Account token. For example, you may need a Service Account token if:
```

daprdocs/content/en/operations/hosting/kubernetes/kubernetes-production.md
```diff
@@ -196,6 +196,20 @@ It is recommended that a production-ready deployment includes the following sett
...
+
+- You are using Dapr components that interact with the Kubernetes APIs, for example the [Kubernetes secret store]({{< ref "kubernetes-secret-store.md" >}}) or the [Kubernetes Events binding]{{< ref "kubernetes-binding.md" >}}).  
+  Note that initializing Dapr components using [component secrets]({{< ref "component-secrets.md" >}}) stored as Kubernetes secrets does **not** require a Service Account token, so you can still set `automountServiceAccountToken: false` in this case. Only calling the Kubernetes secret store at runtime, using the [Secrets management]({{< ref "secrets-overview.md" >}}) building block, is impacted.
```
msfussell, line 208:
```suggestion
{{% alert title="Note" color="primary" %}} 
Initializing Dapr components using [component secrets]({{< ref "component-secrets.md" >}}) stored as Kubernetes secrets does **not** require a Service Account token, so you can still set `automountServiceAccountToken: false` in this case. Only calling the Kubernetes secret store at runtime, using the [Secrets management]({{< ref "secrets-overview.md" >}}) building block, is impacted.
 {{% /alert %}}
```

daprdocs/content/en/operations/hosting/kubernetes/kubernetes-production.md
```diff
@@ -196,6 +196,20 @@ It is recommended that a production-ready deployment includes the following sett
...
+- Your own application needs to interact with the Kubernetes APIs.
+
+Because of the reasons above, Dapr does not set `automountServiceAccountToken: false` automatically for you. However, in all situations where the Service Account is not required by your solution, we do recommend that you set this option in the spec of your pods.
```
msfussell:
```suggestion
Because of the reasons above, Dapr does not set `automountServiceAccountToken: false` automatically for you. However, in all situations where the Service Account is not required by your solution, it is recommended that you set this option in the pods spec.
```

daprdocs/content/en/operations/hosting/kubernetes/kubernetes-production.md
ItalyPaleAle: I don't know if this deserves a large callout? It's more of a * to the sentence above than a full callout in my head (besides, a callout would interrupt the flow of the bulleted list)

hhunter-ms: I agree the shortcode would interrupt the list... if we keep the shortcode, we can move the note under the bullet list?

msfussell: Ok.

hhunter-ms (approved):

PR #3102, opened 2023-01-30
Docs for ServiceInvocationStreaming feature flag

### LINKED ISSUE #3096: [1.10] Docs for ServiceInvocationStreaming feature flag
Document the new feature flag `ServiceInvocationStreaming` and the considerations around streaming.

See: https://github.com/dapr/dapr/pull/5706

ItalyPaleAle:
Fixes #3096

Also removes the Resiliency feature flag from the table


daprdocs/content/en/operations/support/support-preview-features.md
```diff
@@ -14,9 +14,31 @@ For CLI there is no explicit opt-in, just the version that this was first made a
...
-| **Resiliency** | Allows configuring of fine-grained policies for retries, timeouts and circuitbreaking. | `Resiliency` | [Configure Resiliency Policies]({{<ref "resiliency-overview">}}) | v1.7|
 | **App Middleware** | Allow middleware components to be executed when making service-to-service calls | N/A | [App Middleware]({{<ref "middleware.md#app-middleware" >}}) | v1.9 |
+| **Streaming for HTTP service invocation** | Enables (partial) support for using streams in HTTP service invocation; see below for more details. | `ServiceInvocationStreaming` | &nbsp; | v1.10 |
 | **App health checks** | Allows configuring app health checks | `AppHealthCheck` | [App health checks]({{<ref "app-health.md" >}}) | v1.9 |
 | **Pluggable components** | Allows creating self-hosted gRPC-based components written in any language that supports gRPC. The following component APIs are supported: State stores, Pub/sub, Bindings | N/A | [Pluggable components concept]({{<ref "components-concept#pluggable-components" >}})| v1.9  |
+
+### Streaming for HTTP service invocation
```
ItalyPaleAle, line 24: I've added this section to the same page, rather than creating a new page like the other preview features do. The reason is that the content in this section is specifically about the feature flag and how it influences Dapr's behavior, rather than the feature itself, which doesn't require any documentation on its own.

Because of that, it's meant to be temporary until the feature flag is removed too, which makes me feel a bit uncomfortable about creating a TOC item and a URL knowing that it'll be deleted in the future.

Let me know if you think otherwise and I can move this to its own page.

msfussell: I am ok with that, just add a reference link into the documentation column that points to the details below, so people do not miss this.
 "support-preview-features.md#streaming-for-http-service-invocation"

daprdocs/content/en/operations/support/support-preview-features.md
```diff
@@ -14,9 +14,33 @@ For CLI there is no explicit opt-in, just the version that this was first made a
...
+
+Running Dapr with the `ServiceInvocationStreaming` feature flag enables partial support for handling data as a stream in HTTP service invocation. This can offer improvements in performance and memory utilization when using Dapr to invoke another service using HTTP with large request or response bodies.
```
msfussell, line 26: It would be good to clarify this performance improvement. Is this eventually going to be "On" by default so there is always the benefits of streaming and are there some expected perf numbers for improvement (percentage)? It presumably makes impact on these numbers, https://docs.dapr.io/operations/performance-and-scalability/perf-service-invocation/#data-plane-performance which would be useful to reference. 
In other words, why would I turn this on, and we do need to update the perf page above as a result. Interested in hearing your suggestions.

daprdocs/content/en/operations/support/support-preview-features.md
```diff
@@ -14,9 +14,33 @@ For CLI there is no explicit opt-in, just the version that this was first made a
...
+
+The table below summarizes the current state of support for streaming in HTTP service invocation in Dapr, including the impact of enabling `ServiceInvocationStreaming`, in the example where "app A" is invoking "app B" using Dapr. There are six steps in the data flow, with various levels of support for handling data as a stream:
+
+| Step handles data as a stream | Dapr 1.10 | Dapr 1.10 with<br/>`ServiceInvocationStreaming` enabled |
```
msfussell: You may want to pull in this diagram as a reference https://docs.dapr.io/developing-applications/building-blocks/service-invocation/service-invocation-overview/
and then put another column with the numbers. i.e. 1, 3, 4, 5, 6, 7 for clarity.

daprdocs/content/en/operations/support/support-preview-features.md
ItalyPaleAle: Done

daprdocs/content/en/operations/support/support-preview-features.md
ItalyPaleAle: Yes, eventually it will be enabled by default for everyone and will be the only option.

We are still working on perf tests that will prove % improvement, and the challenge is that until the work is complete (which is pending on dapr/dapr#4979 ) , we won't have any way to measure the impact E2E.

The benefit will also be highly variable to each scenario. It will be bigger in cases such as (not exhaustive list):

- When the app is sending a "large" message in the request or response body. "Large" is relative, but I'd say anything over 2KB (the current chunk size) should start seeing some improvement.
- When the producer and/or consumer is slow. This is the case if the data comes from over the network, or if it's read from disk, or if it's generated by a slow producer (in an extreme example, a video encoder)

So it's very much nuanced at this point. Not sure if there's value in explaining the things above?

daprdocs/content/en/operations/support/support-preview-features.md
ItalyPaleAle: Done

daprdocs/content/en/operations/support/support-preview-features.md
```diff
@@ -14,9 +14,35 @@ For CLI there is no explicit opt-in, just the version that this was first made a
 ## Current preview features
 
 | Feature | Description | Setting | Documentation | Version introduced |
-| ---------- |-------------|---------|---------------|-----------------|
+|---|---|---|---|---|
 | **`--image-registry`** flag in Dapr CLI| In self hosted mode you can set this flag to specify any private registry to pull the container images required to install Dapr| N/A | [CLI init command reference]({{<ref "dapr-init.md#self-hosted-environment" >}}) | v1.7 |
-| **Resiliency** | Allows configuring of fine-grained policies for retries, timeouts and circuitbreaking. | `Resiliency` | [Configure Resiliency Policies]({{<ref "resiliency-overview">}}) | v1.7|
 | **App Middleware** | Allow middleware components to be executed when making service-to-service calls | N/A | [App Middleware]({{<ref "middleware.md#app-middleware" >}}) | v1.9 |
+| **Streaming for HTTP service invocation** | Enables (partial) support for using streams in HTTP service invocation; see below for more details. | `ServiceInvocationStreaming` | [Details](#streaming-for-http-service-invocation) | v1.10 |
```
hhunter-ms:
```suggestion
| **Streaming for HTTP service invocation** | Enables (partial) support for using streams in HTTP service invocation; see below for more details. | `ServiceInvocationStreaming` | [Details]({{< ref "support-preview-features.md#streaming-for-http-service-invocation" >}}) | v1.10 |
```

Just bc otherwise it'll break

hhunter-ms (approved):

msfussell (approved):
LGTM

Issue #3091, opened 2023-01-25
Add HTTP binding security token metadata docs


yaron2:
Add content for https://github.com/dapr/components-contrib/pull/2470.

yaron2:
Closed via #3109.

PR #3109, opened 2023-01-31
Add security token for http binding

yaron2:
Closes https://github.com/dapr/docs/issues/3091.

### DIFF
--- a/daprdocs/content/en/reference/components-reference/supported-bindings/http.md
+++ b/daprdocs/content/en/reference/components-reference/supported-bindings/http.md
@@ -21,12 +21,17 @@ spec:
   - name: url
     value: http://something.com
   - name: MTLSRootCA
-    value: /Users/somepath/root.pem #<path to root CA> or <pem encoded string>
+    value: /Users/somepath/root.pem # OPTIONAL <path to root CA> or <pem encoded string>
   - name: MTLSClientCert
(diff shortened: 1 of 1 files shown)


daprdocs/content/en/reference/components-reference/supported-bindings/http.md
```diff
@@ -37,6 +42,8 @@ spec:
 | MTLSRootCA         | N        | Output |Path to root ca certificate or pem encoded string |
 | MTLSClientCert     | N        | Output |Path to client certificate or pem encoded string  |
 | MTLSClientKey      | N        | Output |Path client private key or pem encoded string |
+| securityToken      | N        | Output |The value of a token to be added to an HTTP request as a header. Used together with `securityTokenHeader` |
+| securityTokenHeader| N        | Output |The name of the header for `securityToken` on an HTTP request that |
```
mukundansundar, line 46: Add an example of a header ?
Can see in the above config... Feel it would also be good to be here

mukundansundar (approved):

daprdocs/content/en/reference/components-reference/supported-bindings/http.md
yaron2: In other examples we generally don't put example values in this table, and values are usually displayed if there's a default one specified.. so prefer to keep this as is.

PR #3130, opened 2023-02-02
Pubsub component reference: add aria-label to emojis

ItalyPaleAle:
Small a11y fix (just like #3031)


msfussell (approved):
Good catch. We will pay greater attention to these in the future. Appreciate this.

PR #3150, opened 2023-02-08
Updated resourcepath in statementmanagement quickstart

mlankamp:
## Description

The components folder in the quickstart is renamed to resources, with this change the command line example is changed


mlankamp:
There is an existing PR for this

PR #3179, opened 2023-02-15
Document all brokers support bulk pub/sub

yaron2:
The docs today are wrong in that only Apache Kafka appears to support bulk pub/sub. This is incorrect, as every call to the Dapr APIs, whether through an SDK or directly via HTTP and gRPC, will succeed and accept bulk messages for every pub/sub component. The same is true for subscriptions.

The initial thinking behind the current docs was to show that Kafka has a native driver support for bulk publishing. This is an internal, non-user facing implementation detail that shouldn't give users the wrong impression that Dapr will not accept requests for bulk pub/sub brokers other than Kafka.


msfussell (approved):
LGTM

PR #3194, opened 2023-02-20
Fix dapr run template reference

jjcollinge:
## Description
Fix template reference file environment variables. Currently uses an array when it's decoded into a `map[string]string`

Not sure which branch to target so please advise if v1.11 is not correct.


yaron2:
~Don't we want this in the 1.10 branch?~

Yes, please target the 1.10 branch. Thanks!

jjcollinge:
Ok will update, thanks!

msfussell (approved):
LGTM

PR #3195, opened 2023-02-20
Fix app protocol in template reference doc

jjcollinge:
## Description
Currently the template reference document uses the app protocol strings `HTTP` and `GRPC`, however, when you use these they do not match the strings defined in `daprd` and it fails to setup the app channel. They are currently compared in a case sensitive way which we may want to fix but for now this doc is incorrect. See: https://github.com/dapr/dapr/blob/41dd1dc182c94e40eb132576e1306b5cbe356746/pkg/runtime/runtime.go#L2943


hhunter-ms (approved):
lgtm

PR #3232, opened 2023-03-06
[Resiliency] Clarify timeout section

### LINKED ISSUE #2918: Doc Issue - Resiliency Timeout policy- Default values
In the Resiliency documentation under Time-out policies, https://docs.dapr.io/operations/resiliency/policies/#timeouts - needs below points to be highlighted:

1. Is the time-out value mandatory?
2. what is the behavior if time-out value is not specified
3. what is the max. time-out value can be configured?

hhunter-ms:
## Description

Clarify whether:
- timeout policy is necessary
- behavior if timeout value is not specified
- max timeout value?

## Issue reference

PR will close: #2918

### DIFF
--- a/daprdocs/content/en/operations/resiliency/policies.md
+++ b/daprdocs/content/en/operations/resiliency/policies.md
@@ -12,12 +12,12 @@ Define timeouts, retries, and circuit breaker policies under `policies`. Each po
 
 ## Timeouts
 
-Timeouts can be used to early-terminate long-running operations. If you've exceeded a timeout duration:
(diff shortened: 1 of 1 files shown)


daprdocs/content/en/operations/resiliency/policies.md
```diff
@@ -31,6 +31,8 @@ spec:
       largeResponse: 10s
 ```
 
+If you don't specify a timeout value, the policy will not enforce a time and will default to whatever you set up per the request client.
```
msfussell:
```suggestion
If you don't specify a timeout value, the policy does not enforce a time and defaults to whatever you set up per the request client. 
```

greenie-msft (approved):

PR #3277, opened 2023-03-18
Adds TTL support to SQL server

JoshVanL:
Implemented [here](https://github.com/dapr/components-contrib/pull/2682).

### DIFF
--- a/daprdocs/content/en/reference/components-reference/supported-state-stores/setup-sqlserver.md
+++ b/daprdocs/content/en/reference/components-reference/supported-state-stores/setup-sqlserver.md
@@ -33,6 +33,10 @@ spec:
     value: <SCHEMA> # Optional. defaults to "dbo"
   - name: indexedProperties
     value: <INDEXED-PROPERTIES> # Optional. List of IndexedProperties.
+  - name: metadataTableName # Optional. Name of the table where to store metadata used by Dapr
+    value: "dapr_metadata"
+  - name: cleanupIntervalInSeconds # Optional. Cleanup interval in seconds, to remove expired rows
+    value: 300
 
 ```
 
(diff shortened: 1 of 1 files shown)


daprdocs/content/en/reference/components-reference/supported-state-stores/setup-sqlserver.md
```diff
@@ -80,6 +86,17 @@ When connecting with a dedicated user (not `sa`), these authorizations are requi
 - `CREATE TABLE`
 - `CREATE TYPE`
 
+### TTLs and cleanups
+
+This state store supports [Time-To-Live (TTL)]({{< ref state-store-ttl.md >}}) for records stored with Dapr. When storing data using Dapr, you can set the `ttlInSeconds` metadata property to indicate after how many seconds the data should be considered "expired".
+
+Because SQL Server doesn't have built-in support for TTLs, you implement this in Dapr by adding a column in the state table indicating when the data should be considered "expired". "Expired" records are not returned to the caller, even if they're still physically stored in the database. A background "garbage collector" periodically scans the state table for expired rows and deletes them.
```
ItalyPaleAle:
```suggestion
Because SQL Server doesn't have built-in support for TTLs, Dapr implements this by adding a column in the state table indicating when the data should be considered "expired". "Expired" records are not returned to the caller, even if they're still physically stored in the database. A background "garbage collector" periodically scans the state table for expired rows and deletes them.
```

daprdocs/content/en/reference/components-reference/supported-state-stores/setup-sqlserver.md
```diff
@@ -80,6 +86,17 @@ When connecting with a dedicated user (not `sa`), these authorizations are requi
...
+
+You can set the interval for the deletion of expired records with the `cleanupIntervalInSeconds` metadata property, which defaults to 3600 seconds (that is, 1 hour).
+
+- Longer intervals require less frequent scans for expired rows, but can require storing expired records for longer, potentially requiring more storage space. If you plan to store many records in your state table, with short TTLs, consider setting `cleanupIntervalInSeconds` to a smaller value - for example, `300` (300 seconds, or 5 minutes).
+- If you do not plan to use TTLs with Dapr and the SQL Server state store, you should consider setting `cleanupIntervalInSeconds` to a value <= 0 (e.g. `0` or `-1`) to disable the periodic cleanup and reduce the load on the database.
+
```
ItalyPaleAle, line 99: Can you also add a note saying that the column doesn't have an index, and the consequences of that (garbage collectors perform full-table scans).

Whether an index is required or not depends on the use case. A table with lots of rows (hundreds of thousands) where only some have a TTL could benefit from an index as FTS in this case would be expensive and would remove few rows only. A table with not many rows and/or where most rows have a TTL (like, tables used for caching) would probably benefit from NOT having an index. An index makes queries faster but requires more memory, more storage, and (slightly) slows down inserts.

JoshVanL: Thanks @ItalyPaleAle!

I disagree with this statement though:
`and/or where most rows have a TTL (like, tables used for caching) would probably benefit from NOT having an index`

Surely an index is beneficial because the index orders the `ExpireDate` and therefore exits sooner when performing the `<` condition on the scan?

ItalyPaleAle: Uhmmm it may really depend on how many rows are in the table, and how many need to be deleted on the iteration. I guess we could benchmark that, but I don't care enough 😅 Point is, users may choose to add an index, or not, depending on their specific patterns.

ItalyPaleAle:
@dapr/approvers-docs CC

hhunter-ms (approved):
lgtm

ItalyPaleAle (approved):

PR #3282, opened 2023-03-21
Add priority for rabbitmq

yaron2:
Docs for https://github.com/dapr/components-contrib/pull/2680.


daprdocs/content/en/reference/components-reference/supported-pubsub/setup-rabbitmq.md
```diff
@@ -187,7 +187,7 @@ The routing key is defined by the `routingKey` metadata. For example, if an app
 apiVersion: dapr.io/v1alpha1
 kind: Subscription
 metadata:
-  name: order_pub_sub
+  name: orderspubsub
```
yaron2, line 190: Changed to a name that will work on K8s. an underscore will cause an error

hhunter-ms (approved):

PR #3356, opened 2023-04-28
fix link

hhunter-ms:
fix broken aws link


codefromthecrypt:
can you cherry-pick this into v1.11? I accidentally opened #3360 as v1.11 is broken for the same reason

PR #3394, opened 2023-05-09
v1.10: Fix link to podman installation

JoshVanL:
The current link currently 404s. This changes it to an alternative link.


hhunter-ms:
@JoshVanL can you pull the build fix we pushed into 1.10?

hhunter-ms (approved):

Issue #2766, opened 2022-09-05
Add formats for Actor runtime configuration


shubham1172:
**What content needs to be created or modified?**

Runtime configuration format.

**Describe the solution you'd like**

It is unclear that what is the format of different timeouts and intervals. They use Go's https://pkg.go.dev/time#ParseDuration format. It should be mentioned in the documentation.

**Where should the new material be placed?**

https://v1-9.docs.dapr.io/developing-applications/building-blocks/actors/howto-actors/#actor-runtime-configuration 

**The associated pull request from dapr/dapr, dapr/components-contrib, or other Dapr code repos**

**Additional context**

PR #3439, opened 2023-05-18
[actor runtime config] Add note for timeouts and intervals format

hhunter-ms:
## Description

Add note clarifying that the timeouts and intervals use Go ParseDuration format

## Issue reference

PR will close: #2766


daprdocs/content/en/developing-applications/building-blocks/actors/actors-runtime-config.md
```diff
@@ -19,6 +19,10 @@ You can modify the default Dapr actor runtime behavior using the following confi
 | `remindersStoragePartitions` | Configure the number of partitions for actor's reminders. If not provided, all reminders are saved as a single record in actor's state store. | 0 |
 | `entitiesConfig` | Configure each actor type individually with an array of configurations. Any entity specified in the individual entity configurations must also be specified in the top level `entities` field. | N/A |
 
+{{% alert title="Note" color="primary" %}}
+Timeouts and intervals use [Go's ParseDuration](https://pkg.go.dev/time#ParseDuration) format.
```
shubham1172: Should we include 1-2 examples as well?

shubham1172 (approved):

daprdocs/content/en/developing-applications/building-blocks/actors/actors-runtime-config.md
```diff
@@ -19,6 +19,15 @@ You can modify the default Dapr actor runtime behavior using the following confi
 | `remindersStoragePartitions` | Configure the number of partitions for actor's reminders. If not provided, all reminders are saved as a single record in actor's state store. | 0 |
 | `entitiesConfig` | Configure each actor type individually with an array of configurations. Any entity specified in the individual entity configurations must also be specified in the top level `entities` field. | N/A |
 
+{{% alert title="Note" color="primary" %}}
+Actor timeouts and intervals use [Go's ParseDuration](https://pkg.go.dev/time#ParseDuration) format. You can use various string formats to represent durations, including:
```
msfussell:
```suggestion
Actor timeouts and intervals use [Go's ParseDuration](https://pkg.go.dev/time#ParseDuration) format. You can use various string formats to represent durations, for example:
```

msfussell:
@shubham1172 - Actually is confusing to add this into the docs, since Actors are used through an SDK and this "translates" to the Go format I presume. So for .NET (as the examples show) you only need to use the `TimeSpan` class. So IMO this only needs to be added under the GoLang Tab. Is that correct?

shubham1172:
> @shubham1172 - Actually is confusing to add this into the docs, since Actors are used through an SDK and this "translates" to the Go format I presume. So for .NET (as the examples show) you only need to use the `TimeSpan` class. So IMO this only needs to be added under the GoLang Tab. Is that correct?

@msfussell actually this is not for any SDK, but for the API itself. The examples in the docs are given for various SDKs which handle this internally, but the params above are for the API itself. We could also add it to https://v1-11.docs.dapr.io/reference/api/actors_api/#dapr-calling-to-user-service-code, instead of this place (or both) to avoid any confusions.

msfussell:
> > @shubham1172 - Actually is confusing to add this into the docs, since Actors are used through an SDK and this "translates" to the Go format I presume. So for .NET (as the examples show) you only need to use the `TimeSpan` class. So IMO this only needs to be added under the GoLang Tab. Is that correct?
> 
> @msfussell actually this is not for any SDK, but for the API itself. The examples in the docs are given for various SDKs which handle this internally, but the params above are for the API itself. We could also add it to https://v1-11.docs.dapr.io/reference/api/actors_api/#dapr-calling-to-user-service-code, instead of this place (or both) to avoid any confusions.

Yes, this is a better place and the correct place for this. Can we move this into this topic?

shubham1172:
Hi @hhunter-ms, would you be able to help with moving this doc to API reference instead? Or else I can also do it, please let me know.

hhunter-ms:
> Hi @hhunter-ms, would you be able to help with moving this doc to API reference instead? Or else I can also do it, please let me know.

@shubham1172 Just pushed! am I moving just the note about ParseDuration, or the whole doc?

shubham1172 (approved):

shubham1172:
> > Hi @hhunter-ms, would you be able to help with moving this doc to API reference instead? Or else I can also do it, please let me know.
> 
> @shubham1172 Just pushed! am I moving just the note about ParseDuration, or the whole doc?

Just the note should be fine! LGTM. Thanks @hhunter-ms !!

daprdocs/content/en/reference/api/actors_api.md
```diff
@@ -484,6 +484,16 @@ Parameter | Description
 `maxStackDepth` | A value in the reentrancy configuration that controls how many reentrant calls be made to the same actor.
 `entitiesConfig` | Array of entity configurations that allow per actor type settings. Any configuration defined here must have an entity that maps back into the root level entities.
 
+
+{{% alert title="Note" color="primary" %}}
+Actor timeouts and intervals use [Go's ParseDuration](https://pkg.go.dev/time#ParseDuration) format. You can use various string formats to represent durations. For example:
```
msfussell: To keep this consistent with the Reminders section in the same topic (higher up in the topic), I suggest the following

```suggestion
Actor settings in configuration for timeouts and intervals use [time.ParseDuration](https://pkg.go.dev/time#ParseDuration) format. You can use string formats to represent durations. For example:
```

msfussell: Also I suggest that this Request Body https://github.com/dapr/docs/blob/1018f22d622c4d676a273ccdc989864ce017a4a9/daprdocs/content/en/reference/api/actors_api.md#request-body  section in the Reminders section is also referenced from the Timers section (in this document), since they are the same format. Either cut and paste to the Timers sections or in the Timers section put  link and a comment that this format also applies here. Currently it does not mention the format for Timers

msfussell (commented):
My comments

msfussell (commented):
My comments

daprdocs/content/en/reference/api/actors_api.md
```diff
@@ -484,6 +484,18 @@ Parameter | Description
...
 
+
+{{% alert title="Note" color="primary" %}}
+Actor settings in configuration for timeouts and intervals use [time.ParseDuration](https://pkg.go.dev/time#ParseDuration) format. You can use string formats to represent durations. For example:
+- `1h30m` or `1.5h`: A duration of 1 hour and 30 minutes
+- `1d12h`: A duration of 1 day and 12 hours
+- `500ms`: A duration of 500 milliseconds
+- `-30m`: A negative duration of 30 minutes
+
+This format also applies when [creating actor reminders]({{< ref "#request-body" >}}).
```
msfussell:
```suggestion
```

msfussell (commented):
@hhunter-ms - I updated the doc directly with the link in the Timer section. Please pull down from your branch locally. The link is not needed here.

hhunter-ms:
@msfussell got your changes - lmk if good to merge

msfussell (approved):
LGTM

PR #3459, opened 2023-05-26
V1.11 get started component improvements

JoshVanL:
Currently, the tutorial file paths don't exactly line up if you follow the tutorial to the letter, and leaves a TODO to the reader which is really necessary.

Updates the tutorial to be more copy+pastable.


msfussell (approved):
LGTM. Thanks for the update

PR #3479, opened 2023-06-02
Show same dapr version in quickstart CLI version

JoshVanL:
Show the same version in install command as that in instruction text.


msfussell (approved):
Great catch!

Issue #3485, opened 2023-06-05
Broken link


msfussell:
The link here https://v1-11.docs.dapr.io/reference/components-reference/supported-configuration-stores/ to Postgresql is broken

msfussell:
Fixed

PR #3488, opened 2023-06-05
Rename component spec file to fix link

hhunter-ms:
## Description

'Postgres' was updated to 'PostgreSQL' in the docs, including links, but the actual file name for the PostgreSQL doc wasn't updated.

## Issue reference

PR will close: #3485


msfussell (approved):
This is breaking URL change, but we can live with this.

PR #3518, opened 2023-06-08
Crypto components: use "Using the Dapr cryptography engine" category

ItalyPaleAle:
Renamed the category from "generic" to "Using the Dapr cryptography engine" (open to suggestions as to what the name should be).

This is to differentiate with "generic" components (that we'll have in the future, such as Hashicorp Vault) where the key is not managed by daprd.


msfussell (approved):
LGTM

PR #3524, opened 2023-06-09
Document that `||` is a reserved string that cannot be used in state keys.

JoshVanL:
Found when fuzz testing APIs.


msfussell (approved):
LGTM

PR #3563, opened 2023-06-17
Fix package url for grpc integration

jaeheonji:

## Description

When I followed the example code exactly, I saw an error in the application because of the wrong import path.

I'm not sure about the history, but it seems that the stub code of protobuf has been moved from `dapr/dapr` to `dapr/go-sdk`.
Since `UnimplementedAppCallbackServer` is imported from `dapr/go-sdk`, `common` should also be imported from same package.


daprdocs/content/en/developing-applications/integrations/gRPC-integration.md
```diff
@@ -131,7 +131,7 @@ The following steps will show how to create an app that exposes a server for wit
     	"github.com/golang/protobuf/ptypes/any"
     	"github.com/golang/protobuf/ptypes/empty"
     
-    	commonv1pb "github.com/dapr/dapr/pkg/proto/common/v1"
+    	commonv1pb "github.com/dapr/go-sdk/dapr/proto/common/v1"
```
yaron2: You are right that this needs to change, however the most correct value here would still be `dapr/dapr`. Please change to the following: `commonv1pb "github.com/dapr/dapr/proto/common/v1"`

yaron2 (changes requested):

daprdocs/content/en/developing-applications/integrations/gRPC-integration.md
jaeheonji: when I checked that location, there is only `proto` file. So, I can't get the stub code for go.

I think we have to choose either `dapr/dapr` or `dapr/go-sdk` to unify the package location of runtime and common.

And, as you already mentioned, If `dapr/dapr` is more correct, we should choose this.

yaron2: @jaeheonji yes, we should use `dapr/dapr` all the way in the native gRPC examples. once you make that change we can merge :)

yaron2 (approved):

yaron2:
Thanks for your contribution!

PR #3569, opened 2023-06-20
[Workflow] Make quickstart intro more general

### LINKED ISSUE #3568: [Workflow] Fix quickstart intro
Workflow intro currently lists the .NET workflow activities, which doesn't apply to the Python example
Reword this to make it more general and save SDK specific things for the tabs

hhunter-ms:
## Description

Moved .NET specific workflow activities to .NET tab
Added similar list of workflow activities to Python tab
Update diagram to be more general instead of .NET specific

## Issue reference

PR will close: #3568


msfussell (approved):
LGTM

PR #3609, opened 2023-07-13
Link to related docs from building block index pages

### LINKED ISSUE #1549: Building blocks articles do not link to the SDK topic for that building block
**Describe the issue**
Building block articles today only provide overview and how-to's using base http or rest.  

For example, if I start with the Service Invocation topic, I see Overview and How To invoke using HTTP.  
https://docs.dapr.io/developing-applications/building-blocks/service-invocation/

I do not see the corresponding SDK docs that show a productive way to call using SDKs for each language.  These are hard to find.  Here is the actual link I wanted for Python:
https://docs.dapr.io/developing-applications/sdks/python/python-client/

**Expected content**
I expect a third link type on these building block topics with the SDK links for each language.  Or alternatively we make How-To's include Tabs for HTTP, Python, .NET, Go and so on.  

**Screenshots**
image

**Additional context**

hhunter-ms:
## Description

Seeing if linking to SDK docs, API reference docs, quickstarts, etc from building block index pages works

## Issue reference

PR will close: #1549


msfussell:
This is a great idea and look great! Let's check these in.

msfussell (approved):

PR #3611, opened 2023-07-13
[pub/sub] Clarify component name and add note for SNS/SQS

### LINKED ISSUE #1554: Component name usage isn't clear
**Describe the issue**
The snssqs component yaml is valid but its `metadata.name` attribute is `snssqs-pubsub`
The pubsub building blocks documentation however, is expecting to be using a `pubsub` component name in the default `pubsubname`. Therefore, using the documented component for `snssqs` doesn't work with the pubsub building blocks guide

**URL of the docs**
https://docs.dapr.io/reference/components-reference/supported-pubsub/setup-aws-snssqs/
https://docs.dapr.io/developing-applications/building-blocks/pubsub/howto-publish-subscribe/

**Expected content**
The expected behavior is that the underlying component could be seamlessly replaced between the `redis` pubsub component to the `snssqs` component and work flawlessly with the PubSub tutorial and its deployments/services

**Additional context**
* The above took a while to figure out, as I am a total newbie with respect to Dapr. 
* I was trying to PoC the basic presumption that the component system for PubSub in Dapr is indeed interchangeable and I would be able to use AWS SNS-SQS component and by that to replace the Redis stream component that is described in the PubSub building-blocks tutorial. 
* This wasn't working until I figured out that the problem lies in the fact that the `react-form` (the publisher) and the 2 subscribers are hardcoded to use `pubsubname` value of `pubsub`.
* Another point missing from the PubSub components documentation is the fact that the component is automatically generating the SNS topics and provisions SQS

hhunter-ms:
## Description

Try to clarify docs due to questions around:
- Component name in the how-to - incorrect per RabbitMQ spec
- Add little note about how the SNS/SQS component generates SNS and provisions SQS

## Issue reference

PR will close: #1554


hhunter-ms:
@amimimor could you review?

daprdocs/content/en/reference/components-reference/supported-pubsub/setup-aws-snssqs.md
```diff
@@ -9,7 +9,7 @@ aliases:
 
 ## Component format
 
-To set up AWS SNS/SQS pub/sub, create a component of type `pubsub.aws.snssqs`. See the [pub/sub broker component file]({{< ref setup-pubsub.md >}}) to learn how ConsumerID is automatically generated. Read the [How-to: Publish and Subscribe guide]({{< ref "howto-publish-subscribe.md#step-1-setup-the-pubsub-component" >}}) on how to create and apply a pub/sub configuration.
+To set up AWS SNS/SQS pub/sub, create a component of type `pubsub.aws.snssqs`. The AWS SNS/SQS component automatically generates the SNS topics and provisions SQS.
```
amimimor: I'd write so to emphasize:  `By default, the AWS SNS/SQS .....`. I'm a little concerned with the word "automatically". my proposal is therefore this:
By default, the AWS SNS/SQS component generates the SNS topics, provisions the SQS queues, and configures a subscription of the queues to the topics.
I also don't know if this is the right sentence to detail the fact to without a subscriber (i.e. only publisher that tries to publish a message exists) only the SNS topics are created, whereas if there is a subscriber (with dynamic or static subscription) sns, sqs and the subscription thereof would be generated

hhunter-ms: @amimimor thanks for the review! I've updated - [how is this?](https://github.com/dapr/docs/pull/3611/files#diff-4dd30a3d9cc019d1b443b651b5a7e408d0426b684e8492f160bf10a950679bd3)

amimimor: LGTM!

msfussell (approved):
LGTM

PR #3615, opened 2023-07-14
[AKS integration] Add RBAC as prereq

### LINKED ISSUE #1200: Create documentation for AKS that states RBAC is necessary
I created a K8s cluster that did not have RBAC enabled and the dapr-sidecar-injector was in an infinite (CrashLoopBackOff) loop. The documentation states Dapr will work in any cluster. That is not accurate. Not sure if the answer is to simply update the docs to make it clear that RBAC is required or make Dapr work in any cluster.

hhunter-ms:
## Description

See title

## Issue reference

PR will close: #1200


msfussell (approved):
LGTM

PR #3619, opened 2023-07-18
[bindings] Update `decodeBase64` to input

### LINKED ISSUE #3589: Incorrect documentation on decodeBase64 metadata of Azure Storage Queue input binding
**Describe the issue**
Documentation states that decodeBase64 is used for the Output binding, but after looking at the code it is used in the Input binding related code only (in the Read method) see https://github.com/dapr/components-contrib/blob/master/bindings/azure/storagequeues/storagequeues.go#L169

**URL of the docs**
https://docs.dapr.io/reference/components-reference/supported-bindings/storagequeues/

**Expected content**
decodeBase64 should be marked as Input.
Details should be "Configuration to decode base64 the content received from the Storage Queue into a string"

**Screenshots**

**Additional context**

hhunter-ms:
## Description

See title. Also updated description per issue author.

## Issue reference

PR will close: #3589


msfussell (approved):
LGTM

PR #3640, opened 2023-07-26
Update policies.md

Gcaya:
## Description

Just updated a reference that was I guess outdated.


hhunter-ms (approved):
thanks for catching this!

PR #3674, opened 2023-08-10
Fix CloudEvent override properies

olitomlinson:
## Description

Docs incorrectly use a dash `-` to separate the `cloudevent` and the `field` i.e. `cloudevent-id` it should be `cloudevent.id`


msfussell (approved):
LGTM

PR #3684, opened 2023-08-18
Add memory clarification for components

yaron2:
Let users know it's recommended to set the memory capacity for Dapr for components that read data.

### DIFF
--- a/daprdocs/content/en/concepts/components-concept.md
+++ b/daprdocs/content/en/concepts/components-concept.md
@@ -19,7 +19,9 @@ As another example, the [pub/sub]({{< ref "pubsub-overview.md" >}}) building blo
 
 You can get a list of current components available in the hosting environment using the `dapr components` CLI command.
 
-*Note: For any component that returns data to the app, it is recommended to set the memory capacity of the Dapr process/container accordingly to avoid potential OOM panics*
+{{% alert title="Note" color="primary" %}} 
+For any component that returns data to the app, it is recommended to set the memory capacity of the Dapr process/container accordingly to avoid potential OOM panics.*
+{{% /alert %}}
 
 ## Component specification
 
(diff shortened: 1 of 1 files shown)


daprdocs/content/en/concepts/components-concept.md
```diff
@@ -19,6 +19,8 @@ As another example, the [pub/sub]({{< ref "pubsub-overview.md" >}}) building blo
 
 You can get a list of current components available in the hosting environment using the `dapr components` CLI command.
 
+*Note: for components that read data, it is recommended to set the memory capacity of the Dapr process/container accordingly to avoid potential OOM panics*
```
msfussell:
```suggestion
{{% alert title="Note" color="primary" %}} 
For components that read data, it is recommended to set appropriate memory capacity of the Dapr sidecar (process or container) to avoid potential OOM panics. For example in Docker use the `--memory` option. For process this depend on the OS. 
{{% /alert %}}
```

msfussell: What do you mean specifically by read data here, this is broad. Nearly every component reads data and do you means, reads data into memory? Without guidance this statement is not useful. For example does every pub/sub component read data? At least provide an example.

yaron2: Updated to clarify what reading data means

daprdocs/content/en/concepts/components-concept.md
```diff
@@ -19,6 +19,10 @@ As another example, the [pub/sub]({{< ref "pubsub-overview.md" >}}) building blo
 
 You can get a list of current components available in the hosting environment using the `dapr components` CLI command.
 
+{{% alert title="Note" color="primary" %}} 
+For any component that returns data to the app, it is recommended to set the memory capacity of the Dapr process/container accordingly to avoid potential OOM panics. For example in Docker use the `--memory` option. For Kubernetes, use the `dapr.io/sidecar-memory-limit` annotation. For processes this depends on the OS and/or process orchestration tools.*
```
msfussell:
```suggestion
For any component that returns data to the app, it is recommended to set the memory capacity of the Dapr sidecar accordingly (process or container) to avoid potential OOM panics. For example in Docker use the `--memory` option. For Kubernetes, use the `dapr.io/sidecar-memory-limit` annotation. For processes this depends on the OS and/or process orchestration tools.*
```

msfussell (approved):
LGTM

PR #3721, opened 2023-09-12
Add html for rendering markdown tables

hhunter-ms:
Markdown doesn't render within {{< tabs >}}, so creating a simple html layout file for tables


msfussell (approved):
How is this used? Do you have an example PR using this?

PR #3764, opened 2023-09-25
Example in documentation is not correct

Kuzbekov:
## Description

field ".spec.version" does not exist in httpendpoint declaration


yaron2 (approved):

yaron2:
Thanks for your contribution!

PR #3582, opened 2023-06-27
Adds Etcdv2

JoshVanL:
See https://github.com/dapr/components-contrib/pull/2932

PR #3809, opened 2023-10-11
Update etcd version in table

ItalyPaleAle:
See #3582


daprdocs/data/components/state_stores/generic.yaml
```diff
@@ -45,7 +45,7 @@
 - component: etcd
   link: setup-etcd
   state: Beta
-  version: v1
+  version: v2
   since: "1.11"
```
msfussell: Should change this to indicate when this version was available from
```suggestion
  since: "1.12"
```

yaron2 (approved):

PR #3813, opened 2023-10-11
Update docs for routeralias middleware

### LINKED ISSUE #3525: [1.12] Update metadata for router alias middleware
The metadata for the component is changing in 1.12 to make it more consistent with the rest of Dapr.

Rather than passing each route as a metadata key-value pair, there's a single metadata property `routes` which is a JSON-encoded or YAML-encoded string with the key-value pairs of the routes.

See: dapr/components-contrib#2903

ItalyPaleAle:
Fixes #3525


mukundansundar (approved):

hhunter-ms (approved):

msfussell (approved):
LGTM

PR #3814, opened 2023-10-11
Remove direction metadata from output-only bindings

ItalyPaleAle:
Setting the `direction` metadata on output-only binding is a no-op. This removes that property from the metadata of bindings that only implement the output binding interface to avoid confusing users.

The metadata.yaml generator already handles this case, not listing the "direction" property for bindings that are only output bindings: https://github.com/dapr/components-contrib/blob/9f2cc5c158bbdf407b35d16ff9a2d79a172dcd83/.build-tools/pkg/metadataschema/validators.go#L203


msfussell:
Agreed. It does make sense to only show this on input bindings for optimization.

msfussell (approved):
LGTM

hhunter-ms (approved):

PR #3850, opened 2023-10-25
[Memcached] Remove 'ms' from `timeout` example

### LINKED ISSUE #2820: Memcached `timeout` value is numeric not String
**Describe the issue**

Memcache State Store documentation states that the default value for its `timeout` field is `1000ms`. This leads the user to think that you can suffix a number with "ms" (milliseconds) or "s" (for seconds) sand that Dapr will make sense of it. But that is not the case That field expects a number in a unit of milliseconds. Any string content will be ignored. 

**URL of the docs**

* https://docs.dapr.io/reference/components-reference/supported-state-stores/setup-memcached/

**Expected content**

Make it clear that the unit used by the field is miliseconds.

**Screenshots**

**Additional context**

hhunter-ms:
## Description

Timeout is only an integer, so 'ms' will be ignored. Removing from examples and clarifying that it is in milliseconds in table.

## Issue reference

PR will close: #2820


hhunter-ms:
@tmacam how is this?

yaron2 (approved):

Issue #2763, opened 2022-09-01
Update AWS SSM Parameter Store secret store docs


ostreek:
**What content needs to be created or modified?**
The documentation for the AWS SSM Parameter Store secret store needs to be updated for https://github.com/dapr/components-contrib/issues/2001. I also noticed there's functionality to retrieve a specific version of a parameter, but this isn't documented.

**Describe the solution you'd like**
The prefix option needs to be documented. Ideally the parameter store docs would have an example section like Azure Key Vault has.

**Where should the new material be placed?**
In the [AWS SSM Parameter Store component spec](https://docs.dapr.io/reference/components-reference/supported-secret-stores/aws-parameter-store/).
The option to retrieve a specific version of a parameter should also be mentioned in the [API reference](https://docs.dapr.io/reference/api/secrets_api/#query-parameters).

PR #3920, opened 2023-12-21
[AWS SSM parameter store] Add prefix option

hhunter-ms:
#2763


msfussell:
@berndverst - Can you check this PR for correctness also?

msfussell (approved):
LGTM

hhunter-ms:
@ostreek could you review as well?

ostreek:
Sorry for the slow reply Hannah, I didn't get a notification. Thanks for picking this up. What you've written is correct, I agree with this change.

The docs for the Azure key vault secret store are very good, they give an example of how to create a key vault and configure the component to use it. It'd be great if we could provide similar instructions here, to show how to use the prefix option and explain how it maps to the parameter store. But I suggest you merge this PR now and maybe we can further improve the page later.

Thanks again for documenting this, I'm sorry I didn't do this at the time!

PR #3931, opened 2024-01-01
Describing Kafka consumer behaviour when consuming from multiple topics

### LINKED ISSUE #3707: Kafka Pub/Sub documentation should note consumer group behaviour when subscribing to multiple topics
**Describe the issue**
When using Dapr to subscribe to multiple topics in Kafka, it should be noted on the documentation that the consumer group which is created will be distributed across all the topics. 

Say, for instance, you are subscribing to two topics: one with 10 partitions, and another with 30. You may presume, in order to scale to your absolute maximum for performance, you would want to run 40 pods; with 10 pods being assigned to the first topic, 30 to the other. However, this is not always the case (but could be). There is no guarantee about how consumers are balanced a cross the topic.

The consumer group will do its best to balance consumers, but it can't know for sure what the balance should be. This becomes particularly challenging if you have fairly peaky volumes traversing through your topics, and you use some kind of autoscaler to attempt to match the load, you might end up with pods not consuming and consumer lag growing.

The solution to my problem has been to configure multiple Pub/Sub components, with different names, different consumer groups, all pointing to the same Kafka broker. Ultimately it comes down to spinning up separate consumer groups. Using the default consumer group of `sarama` or even setting the consumer group name to `{appID}` (and possibly `{namespace}` if you only have one namespace) will replicate this problem. For example, my two Pub/Sub components are configured as follows:

```yaml
- name: consumerGroup
  value: "{appID}-topic-one"
- name: clientID 
  value: "{appID}"
```

```yaml
- name: consumerGroup
  value: "{appID}-topic-two"
- name: clientID 
  value: "{appID}"
```

**URL of the docs**
https://docs.dapr.io/reference/components-reference/supported-pubsub/setup-apache-kafka/

**Expected content**
The docs should mention this behaviour and inform users of ways to get around this particular problem. The simplest being multiple Pub/Sub components. This effectively creates two connections to the broker, under separate consumer groups, distributing pods to consumers exactly 1:1.

**Screenshots**
N/A

**Additional context**
N/A

KrylixZA:

## Description

This pull request documents the behaviour of Dapr consumer groups when consuming from multiple topics in Kafka. This updates the documentation to help users understand consumer group behaviour when subscribing to multiple topics to avoid the pitfalls of unevenly balanced consumer groups across topics.

## Issue reference

Closes #3707.


KrylixZA:
In light of the changes made in Dapr 1.12.3, this may not be necessary anymore. We should confirm the behaviour of a single pub/sub component being used to subscribe to multiple topics with Dapr 1.12.3 before either merging or abandoning this pull request.

KrylixZA:
Looking at the code changes in the [fix Kafka consumer](https://github.com/dapr/components-contrib/pull/3299) pull request on the components-contrib repo, my gut feel says the documentation herein is still relevant. Will confirm next week 👍

daprdocs/content/en/reference/components-reference/supported-pubsub/setup-apache-kafka.md
```diff
@@ -299,6 +299,40 @@ auth:
   secretStore: <SECRET_STORE_NAME>
 ```
 
+## Consuming from multiple topics
+
+When consuming from multiple topics using a single pub/sub component, there is no guarantee about how the consumers in your consumer group are balanced across the topic partitions. For instance, if you are subscribing to two topics with 10 partitions per topic and you have 20 replicas of your service consuming from the two topics, then there is no guarantee that 10 will be assigned to the first topic and 10 to the second topic. It could be possible that more than 10 are assigned to the first topic and subsequently less than 10 assigned to the second topic. This can result in idle consumers listening to the first topic and over-extended consumers on the second topic, or vice versa. This same beheaviour can be observed when using auto-scalers such as HPA or KEDA.
+
+If you run into this particular issue, it is recommended that you configure a single pub/sub component per topic with uniquely defined consumer groups per component. This will guarantee that all replicas of your service are fully allocated to the unique consumer group, where each consumer group targets one specific topic.
```
msfussell:
```suggestion
If you run into this particular issue, it is recommended that you configure a single pub/sub component per topic with uniquely defined consumer groups per component. This guarantees that all replicas of your service are fully allocated to the unique consumer group, where each consumer group targets one specific topic.
```

daprdocs/content/en/reference/components-reference/supported-pubsub/setup-apache-kafka.md
```diff
@@ -299,6 +299,40 @@ auth:
   secretStore: <SECRET_STORE_NAME>
 ```
 
+## Consuming from multiple topics
```
msfussell, line 302: Is it best to call this title "Optimizing when consuming multiple topics" or similar?

msfussell (commented):
Thanks for this contribution. Yes this should remain true even after the v1.12.3 release, but best if you can confirm anyway.

daprdocs/content/en/reference/components-reference/supported-pubsub/setup-apache-kafka.md
KrylixZA: Initially I had put this down as "Optimizing" but based on the feedback on the open issue around this behaviour as well as the conversations I've had with other developers, this feels/is like a workaround and not an optimisation. Because of that I opted not to put anything other a simple title, as listed above.

Thoughts?

daprdocs/content/en/reference/components-reference/supported-pubsub/setup-apache-kafka.md
KrylixZA: Suggestion applied. Thanks @msfussell 👍

KrylixZA:
> Thanks for this contribution. Yes this should remain true even after the v1.12.3 release, but best if you can confirm anyway.

Hey Mark. Confirmed the updated documentation in this PR is still relevant in 1.12.3. Teams still need to configure multiple consumer groups (pub/sub component definitions) in order to effectively consume all partitions with all running pods 👍 

This PR can merge when you are happy with it.

daprdocs/content/en/reference/components-reference/supported-pubsub/setup-apache-kafka.md
msfussell: Sounds good. Then let's leave it as is.

msfussell (approved):
Approved. Appreciate the great contribution to the docs

daprdocs/content/en/reference/components-reference/supported-pubsub/setup-apache-kafka.md
```diff
@@ -299,6 +299,40 @@ auth:
...
+
+When consuming from multiple topics using a single pub/sub component, there is no guarantee about how the consumers in your consumer group are balanced across the topic partitions. For instance, if you are subscribing to two topics with 10 partitions per topic and you have 20 replicas of your service consuming from the two topics, then there is no guarantee that 10 will be assigned to the first topic and 10 to the second topic. It could be possible that more than 10 are assigned to the first topic and subsequently less than 10 assigned to the second topic. This can result in idle consumers listening to the first topic and over-extended consumers on the second topic, or vice versa. This same beheaviour can be observed when using auto-scalers such as HPA or KEDA.
```
hhunter-ms:
```suggestion
When consuming from multiple topics using a single pub/sub component, there is no guarantee about how the consumers in your consumer group are balanced across the topic partitions. 

For instance, let's say you are subscribing to two topics with 10 partitions per topic and you have 20 replicas of your service consuming from the two topics. There is no guarantee that 10 will be assigned to the first topic and 10 to the second topic. Instead, the partitions could be divided unequally, with more than 10 assigned to the first topic and the rest assigned to the second topic. 

This can result in idle consumers listening to the first topic and over-extended consumers on the second topic, or vice versa. This same behavior can be observed when using auto-scalers such as HPA or KEDA.
```

broke up the paragraph a bit, just because it makes it easier for people to read/skim

hhunter-ms (changes requested):
quick review

hhunter-ms (approved):

KrylixZA:
Hey @msfussell / @hhunter-ms. Just an FYI: I am not authorised to merge PRs. When you are all happy and ready for this to merge, please would you mind completing the PR?


Ty :)

Issue #3965, opened 2024-01-23
v1.13 docs release endgame


msfussell:
- [x] Do an upmerge from v1.11 to v1.12 @hhunter-ms 
- [x] Do an upmerge from v1.12 to v1.13 @hhunter-ms 

 ## Update all SDKs docs
- [x] All SDK maintainers ACK that docs are ready  
- [x] Update all the SDK submodules and triage changes with maintainers  

## v1.12 Branch - use this for reference:  **** https://github.com/dapr/docs/pull/3166

@hhunter-ms 
In pending PR:
- [x] Add v1.13 (latest) to nav bar
- [x] Archive notice
- [x] Update root README.md
- [x] Update URL in config.toml
- [x] Delete workflow file

Day of:
- [x] Push v1.12 PR @msfussell

## v1.13 Branch - use this for reference: **** https://github.com/dapr/docs/pull/3168

@hhunter-ms
In a pending PR:
- [x] Update README.md
- [x] Update version dropdown in nav bar
- [x] Supported versions section
- [x] All output logs change 1.12 to 1.13 (change dapr-latest-version.html)
- [x] Update URL in config.toml
- [x] Update workflow for docs.dapr.io to v1.13 branch

Day of: 
- [x] Push v1.13 PR @msfussell  
- [x] Update diagram deck @hhunter-ms 

## GitHub - After release

- [x] Make v1.13 the default branch @msfussell  
- [x] One last upmerge @hhunter-ms 

## Website

- [x] Create new website for v1.14 @greenie-msft 
- [x] Add DNS entry for new website @greenie-msft 
- [x] Add custom domain & generate SSL certificate - Needs to happen after the DNS entry is created @greenie-msft 

## v1.14 Branch 

- [x] Create v1.14 branch @hhunter-ms 
- [x] Update website workflow file - See this PR https://github.com/dapr/docs/pull/3178 for the Nav bar @hhunter-ms 
- [x] Upmerge from release branch to preview branch
- [x] Check if need to remove website-root.yml to keep from competing with current branch

## Release announcements

- [x] Release notes/blog @marcduiker 
- [x] Tweet release announcement @marcduiker 
- [x] Discord release announcement @marcduiker

greenie-msft:
Website for 1.13 and 1.14 are ready to go.

1.15 and beyond still need to be created/configured wtih the proper cnames.

msfussell:
v1.13 docs released!

PR #3976, opened 2024-01-26
Upmerge 1.12 --> 1.13

hhunter-ms:
#3965


msfussell (approved):
Should we include this file - daprdocs/layouts/shortcodes/dapr-latest-version.html
Or fix this up to be v1.13.0 afterwards?

hhunter-ms:
> Should we include this file - daprdocs/layouts/shortcodes/dapr-latest-version.html Or fix this up to be v1.13.0 afterwards?

I think we typically update this in one of the other PRs in the endgame issue

PR #3995, opened 2024-01-31
[1.13] Update docs for low-cardinality metrics

### LINKED ISSUE #3987: Explain what the impact is of reducing cardinality metrics
Tracks this issue https://github.com/dapr/dapr/issues/6723
Document the impact of this change on the metrics affected and if a feature flag has been added document its usage and effect.

ItalyPaleAle:
Fixes #3987

Includes documenting the new "http.increasedCardinality" configuration option

### DIFF
--- a/daprdocs/content/en/operations/configuration/configuration-overview.md
+++ b/daprdocs/content/en/operations/configuration/configuration-overview.md
@@ -106,21 +106,27 @@ The `metrics` section under the `Configuration` spec contains the following prop
 ```yml
 metrics:
   enabled: true
+  rules: []
+  http:
+    increasedCardinality: true
 ```
 
 The following table lists the properties for metrics:
 
 | Property     | Type   | Description |
 |--------------|--------|-------------|
-| `enabled` | boolean | Whether metrics should to be enabled. |
-| `rules`   | boolean | Named rule to filter metrics. Each rule contains a set of `labels` to filter on and a`regex`expression to apply to the metrics path. |
+| `enabled` | boolean | When set to true, the default, enables metrics collection and the metrics endpoint. |
+| `rules`   | array | Named rule to filter metrics. Each rule contains a set of `labels` to filter on and a`regex`expression to apply to the metrics path. |
+| `http.increasedCardinality` | boolean | When set to true, in the Dapr HTTP server each request path causes the creation of a new "bucket" of metrics. This can cause issues, including excessive memory consumption, when there many different requested endpoints (such as when interacting with RESTful APIs).<br>In Dapr 1.13 the default value is `true` (to preserve the behavior of Dapr <= 1.12), but will change to `false` in Dapr 1.14. |
 
-To mitigate high memory usage and egress costs associated with [high cardinality metrics]({{< ref "metrics-overview.md#high-cardinality-metrics" >}}), you can set regular expressions for every metric exposed by the Dapr sidecar. For example:
+To mitigate high memory usage and egress costs associated with [high cardinality metrics]({{< ref "metrics-overview.md#high-cardinality-metrics" >}}) with the HTTP server, you should set the `metrics.http.increasedCardinality` property to `false`.
+
+Using rules, you can set regular expressions for every metric exposed by the Dapr sidecar. For example:
 
 ```yml
-metric:
-    enabled: true
-    rules:
+metrics:
+  enabled: true
+  rules:
     - name: dapr_runtime_service_invocation_req_sent_total
       labels:
       - name: method
--- a/daprdocs/content/en/operations/configuration/configuration-overview.md
+++ b/daprdocs/content/en/operations/configuration/configuration-overview.md
@@ -116,7 +116,7 @@ The following table lists the properties for metrics:
 | Property     | Type   | Description |
 |--------------|--------|-------------|
 | `enabled` | boolean | When set to true, the default, enables metrics collection and the metrics endpoint. |
-| `rules`   | array | Named rule to filter metrics. Each rule contains a set of `labels` to filter on and a`regex`expression to apply to the metrics path. |
+| `rules`   | array | Named rule to filter metrics. Each rule contains a set of `labels` to filter on and a `regex` expression to apply to the metrics path. |
 | `http.increasedCardinality` | boolean | When set to true, in the Dapr HTTP server each request path causes the creation of a new "bucket" of metrics. This can cause issues, including excessive memory consumption, when there many different requested endpoints (such as when interacting with RESTful APIs).<br>In Dapr 1.13 the default value is `true` (to preserve the behavior of Dapr <= 1.12), but will change to `false` in Dapr 1.14. |
 
 To mitigate high memory usage and egress costs associated with [high cardinality metrics]({{< ref "metrics-overview.md#high-cardinality-metrics" >}}) with the HTTP server, you should set the `metrics.http.increasedCardinality` property to `false`.
--- a/daprdocs/content/en/operations/observability/metrics/metrics-overview.md
+++ b/daprdocs/content/en/operations/observability/metrics/metrics-overview.md
@@ -10,7 +10,7 @@ By default, each Dapr system process emits Go runtime/process metrics and has th
 
 ## Prometheus endpoint
 
-The Dapr sidecars exposes a [Prometheus](https://prometheus.io/)-compatible metrics endpoint that you can scrape to gain a greater understanding of how Dapr is behaving.
+The Dapr sidecar exposes a [Prometheus](https://prometheus.io/)-compatible metrics endpoint that you can scrape to gain a greater understanding of how Dapr is behaving.
 
 ## Configuring metrics using the CLI
 
--- a/daprdocs/content/en/operations/observability/metrics/metrics-overview.md
+++ b/daprdocs/content/en/operations/observability/metrics/metrics-overview.md
@@ -103,7 +103,7 @@ spec:
             "orders/": "orders/.+"
 ```
 
-When this configuration is applied, a recorded metric with the `method` label of `orders/a746dhsk293972nz` will be replaced with `orders/`.
+When this configuration is applied, a recorded metric with the `method` label of `orders/a746dhsk293972nz` is replaced with `orders/`.
 
 Using regular expressions to reduce metrics cardinality is considered legacy. We encourage all users to set `spec.metrics.http.increasedCardinality` to `false` instead, which is simpler to configure and offers better performance.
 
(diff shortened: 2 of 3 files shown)


daprdocs/content/en/operations/observability/metrics/metrics-overview.md
```diff
@@ -96,12 +105,7 @@ spec:
 
 When this configuration is applied, a recorded metric with the `method` label of `orders/a746dhsk293972nz` will be replaced with `orders/`.
 
-### Watch the demo
```
ItalyPaleAle, line 99: Video was removed as users should leverage the new configuration option rather than create a (very complicated) set of regular expressions

daprdocs/content/en/operations/configuration/configuration-overview.md
```diff
@@ -106,21 +106,27 @@ The `metrics` section under the `Configuration` spec contains the following prop
...
 ```
 
 The following table lists the properties for metrics:
 
 | Property     | Type   | Description |
 |--------------|--------|-------------|
-| `enabled` | boolean | Whether metrics should to be enabled. |
-| `rules`   | boolean | Named rule to filter metrics. Each rule contains a set of `labels` to filter on and a`regex`expression to apply to the metrics path. |
+| `enabled` | boolean | When set to true, the default, enables metrics collection and the metrics endpoint. |
+| `rules`   | array | Named rule to filter metrics. Each rule contains a set of `labels` to filter on and a`regex`expression to apply to the metrics path. |
```
msfussell: The name rule here does not need to match a specific metric name, correct? However it would be good to explain this with at least a couple of examples for some metrics.

daprdocs/content/en/operations/observability/metrics/metrics-overview.md
```diff
@@ -84,9 +92,10 @@ kind: Configuration
...
   name: daprConfig
 spec:
-  metric:
-      enabled: true
-      rules:
+  metrics:
+    enabled: true
+    increasedCardinality: true
+    rules:
       - name: dapr_runtime_service_invocation_req_sent_total
```
msfussell, line 99: - Does dapr_runtime_service_invocation_req_sent_total this need to match a metric name? Presumably not.

daprdocs/content/en/operations/observability/metrics/metrics-overview.md
```diff
@@ -66,15 +66,23 @@ metadata:
...
   metrics:
     enabled: false
 ```
 
 ## High cardinality metrics
 
-Depending on your use case, some metrics emitted by Dapr might contain values that have a high cardinality. This might cause increased memory usage for the Dapr process/container and incur expensive egress costs in certain cloud environments. To mitigate this issue, you can set regular expressions for every metric exposed by the Dapr sidecar. [See a list of all Dapr metrics](https://github.com/dapr/dapr/blob/master/docs/development/dapr-metrics.md).
+When invoking Dapr using HTTP, the legacy behavior (and current default as of Dapr 1.13) is to create a separate "bucket" for each requested method. When working with RESTful APIs, this can cause very high cardinality, with potential negative impact on memory usage and CPU.
+
+Dapr 1.13 introduces a new option for the Dapr Configuration resource `spec.metrics.http.increasedCardinality`: when set to `false`, it reports metrics for the HTTP server for each "abstract" method (for example, requesting from a state store) instead of creating a "bucket" for each concrete request path.
```
msfussell, line 77: Would anyone choose to set this to true for a reason then?

daprdocs/content/en/operations/observability/metrics/metrics-overview.md
```diff
@@ -96,12 +105,7 @@ spec:
 
 When this configuration is applied, a recorded metric with the `method` label of `orders/a746dhsk293972nz` will be replaced with `orders/`.
```
msfussell:
```suggestion
When this configuration is applied, a recorded metric with the `method` label of `orders/a746dhsk293972nz` is replaced with `orders/`.
```

daprdocs/content/en/operations/observability/metrics/metrics-overview.md
msfussell: Does this new option completely remove the need for regular expressions?

msfussell (commented):
Looks good. Could you also add `metrics` into the schema for configuration here https://docs.dapr.io/reference/resource-specs/configuration-schema/#sidecar-format since it is missing currently.

daprdocs/content/en/operations/configuration/configuration-overview.md
ItalyPaleAle: This is not something new -  and TBH it's not something we should promote much. There are some examples below if needed.

daprdocs/content/en/operations/observability/metrics/metrics-overview.md
ItalyPaleAle: Yes that's the metric name. However, here too, this is something we shouldn't be promoting much as it's a legacy thing.

daprdocs/content/en/operations/observability/metrics/metrics-overview.md
ItalyPaleAle: I do wonder that myself!

ItalyPaleAle:
> Could you also add `metrics` into the schema for configuration here https://docs.dapr.io/reference/resource-specs/configuration-schema/#sidecar-format since it is missing currently.

Yes, I'll do that tomorrow

> Does this new option completely remove the need for regular expressions?

Doesn't remove the need entirely, but originally regexs for metrics were "marketed" as a solution to the high cardinality problem. I suspect a lot of users who figured that out (and I can tell you by first-hand experience we did not make it easy to figure that out at all!) were using them for this purpose. However, there could be other purposes and those could still be valid.

daprdocs/content/en/operations/configuration/configuration-overview.md
hhunter-ms:
```suggestion
| `rules`   | array | Named rule to filter metrics. Each rule contains a set of `labels` to filter on and a `regex` expression to apply to the metrics path. |
```

daprdocs/content/en/operations/observability/metrics/metrics-overview.md
```diff
@@ -3,22 +3,24 @@ type: docs
...
-description: "Enable or disable Dapr metrics "
+description: "Enable or disable Dapr metrics"
 ---
 
 By default, each Dapr system process emits Go runtime/process metrics and has their own [Dapr metrics](https://github.com/dapr/dapr/blob/master/docs/development/dapr-metrics.md).
 
 ## Prometheus endpoint
-The Dapr sidecars exposes a [Prometheus](https://prometheus.io/) metrics endpoint that you can scrape to gain a greater understanding of how Dapr is behaving.
+
+The Dapr sidecars exposes a [Prometheus](https://prometheus.io/)-compatible metrics endpoint that you can scrape to gain a greater understanding of how Dapr is behaving.
```
hhunter-ms:
```suggestion
The Dapr sidecar exposes a [Prometheus](https://prometheus.io/)-compatible metrics endpoint that you can scrape to gain a greater understanding of how Dapr is behaving.
```

daprdocs/content/en/operations/observability/metrics/metrics-overview.md
```diff
@@ -66,15 +66,23 @@ metadata:
...
+When invoking Dapr using HTTP, the legacy behavior (and current default as of Dapr 1.13) is to create a separate "bucket" for each requested method. When working with RESTful APIs, this can cause very high cardinality, with potential negative impact on memory usage and CPU.
+
+Dapr 1.13 introduces a new option for the Dapr Configuration resource `spec.metrics.http.increasedCardinality`: when set to `false`, it reports metrics for the HTTP server for each "abstract" method (for example, requesting from a state store) instead of creating a "bucket" for each concrete request path.
+
+The default value of `spec.metrics.http.increasedCardinality` is `true` in Dapr 1.13, to maintain the same behavior as Dapr 1.12 and older. However, the value will change to `false` (low-cardinality metrics by default) in Dapr 1.14.
+
+Setting `spec.metrics.http.increasedCardinality` to `false` is **recommended** to all Dapr users, to reduce resource consumption. The pre-1.13 behavior, which is used when the option is `true`, is considered legacy and is only maintained for users who have special requirements around backwards-compatibility.
```
hhunter-ms, line 81: we can encapsulate this in an 'important' alert if you feel necessary - up to you

```suggestion
{{% alert title="Important" color="warning" %}}
Setting `spec.metrics.http.increasedCardinality` to `false` is **recommended** to all Dapr users, to reduce resource consumption. The pre-1.13 behavior, which is used when the option is `true`, is considered legacy and is only maintained for users who have special requirements around backwards-compatibility.
{{% /alert %}}
```

hhunter-ms (changes requested):
quick review

daprdocs/content/en/operations/observability/metrics/metrics-overview.md
ItalyPaleAle: My .02: since this is something temporary for 1.13 only, I don't think it needs that much highlighting. But I'm not against it either

daprdocs/content/en/operations/observability/metrics/metrics-overview.md
msfussell: But does it need to name match with the metric name or not?

ItalyPaleAle: I *think* so, according to the docs. I'm not super familiar with that feature myself

daprdocs/content/en/operations/observability/metrics/metrics-overview.md
msfussell: @ItalyPaleAle - Any reply to this question?

daprdocs/content/en/operations/observability/metrics/metrics-overview.md
msfussell: Can I suggest then that we add this, which if used, is a good practice anyway. That the name needs to match the metrics name explicitly.

daprdocs/content/en/operations/observability/metrics/metrics-overview.md
hhunter-ms: @ItalyPaleAle that works for me

hhunter-ms (approved):

daprdocs/content/en/operations/observability/metrics/metrics-overview.md
```diff
@@ -66,15 +66,23 @@ metadata:
...
+
+## Transform metrics with regular expressions
+
+You can set regular expressions for every metric exposed by the Dapr sidecar to "transform" their values. [See a list of all Dapr metrics](https://github.com/dapr/dapr/blob/master/docs/development/dapr-metrics.md).
 
 The following example shows how to apply a regular expression for the label `method` in the metric `dapr_runtime_service_invocation_req_sent_total`:
```
msfussell:
```suggestion
The name of the rule must match the name of the metric that is transformed. The following example shows how to apply a regular expression for the label `method` in the metric `dapr_runtime_service_invocation_req_sent_total`:
```

msfussell (approved):
LGTM!

PR #3996, opened 2024-01-31
[1.13] Docs for PostgreSQL v2 state store

### LINKED ISSUE #3911: [1.13] Docs for Postgres state store v2
Need docs for the Postgres state store v2: https://github.com/dapr/components-contrib/pull/3250

Compared to v1:

- Works with CockroachDB, YugabyteDB, with the same code
- Does NOT support query API

V1 will not be deprecated and will remain as an option

ItalyPaleAle:
Fixes #3911

### DIFF
--- a/daprdocs/content/en/reference/components-reference/supported-state-stores/setup-postgresql-v1.md
+++ b/daprdocs/content/en/reference/components-reference/supported-state-stores/setup-postgresql-v1.md
@@ -110,7 +110,7 @@ Authenticating with Microsoft Entra ID is supported with Azure Database for Post
 
      > This example does not describe a production configuration because it sets the password in plain text and the user name is left as the PostgreSQL default of "postgres".
 
-2. Create a database for state data.  
+1. Create a database for state data.  
     Either the default "postgres" database can be used, or create a new database for storing state data.
 
     To create a new database in PostgreSQL, run the following SQL command:
--- a/daprdocs/content/en/reference/components-reference/supported-state-stores/setup-postgresql-v1.md
+++ b/daprdocs/content/en/reference/components-reference/supported-state-stores/setup-postgresql-v1.md
@@ -1,13 +1,23 @@
 ---
 type: docs
-title: "PostgreSQL"
-linkTitle: "PostgreSQL"
-description: Detailed information on the PostgreSQL state store component
+title: "PostgreSQL v1"
+linkTitle: "PostgreSQL v1"
+description: Detailed information on the PostgreSQL v1 state store component
 aliases:
   - "/operations/components/setup-state-store/supported-state-stores/setup-postgresql/"
+  - "/operations/components/setup-state-store/supported-state-stores/setup-postgres/"
+  - "/operations/components/setup-state-store/supported-state-stores/setup-postgresql-v1/"
+  - "/operations/components/setup-state-store/supported-state-stores/setup-postgres-v1/"
 ---
 
-This component allows using PostgreSQL (Postgres) as state store for Dapr. See [this guide]({{< ref "howto-get-save-state.md#step-1-setup-a-state-store" >}}) on how to create and apply a state store configuration.
+{{% alert title="Note" color="primary" %}}
+Starting with Dapr 1.13, you can leverage the [PostgreSQL v2]({{< ref setup-postgresql-v2.md >}}) state store component, which contains some improvements to performance and reliability.  
+The v2 component is not compatible with v1, and data cannot be migrated between the two components. The v2 component does not offer support for state store query APIs.
+
+There are no plans to deprecate the v1 component.
+{{% /alert %}}
+
+This component allows using PostgreSQL (Postgres) as state store for Dapr, using the "v1" component. See [this guide]({{< ref "howto-get-save-state.md#step-1-setup-a-state-store" >}}) on how to create and apply a state store configuration.
 
 ```yaml
 apiVersion: dapr.io/v1alpha1
@@ -21,8 +31,8 @@ spec:
     # Connection string
     - name: connectionString
       value: "<CONNECTION STRING>"
-    # Timeout for database operations, in seconds (optional)
-    #- name: timeoutInSeconds
+    # Timeout for database operations, as a Go duration or number of seconds (optional)
+    #- name: timeout
     #  value: 20
     # Name of the table where to store the state (optional)
     #- name: tableName
@@ -31,8 +41,8 @@ spec:
     #- name: metadataTableName
     #  value: "dapr_metadata"
     # Cleanup interval in seconds, to remove expired rows (optional)
-    #- name: cleanupIntervalInSeconds
-    #  value: 3600
+    #- name: cleanupInterval
+    #  value: "1h"
     # Maximum number of connections pooled by this component (optional)
     #- name: maxConns
     #  value: 0
@@ -59,7 +69,7 @@ The following metadata options are **required** to authenticate using a PostgreS
 
 | Field  | Required | Details | Example |
 |--------|:--------:|---------|---------|
-| `connectionString` | Y | The connection string for the PostgreSQL database. See the PostgreSQL [documentation on database connections](https://www.postgresql.org/docs/current/libpq-connect.html) for information on how to define a connection string. | `"host=localhost user=postgres password=example port=5432 connect_timeout=10 database=my_db"`
+| `connectionString` | Y | The connection string for the PostgreSQL database. See the PostgreSQL [documentation on database connections](https://www.postgresql.org/docs/current/libpq-connect.html) for information on how to define a connection string. | `"host=localhost user=postgres password=example port=5432 connect_timeout=10 database=my_db"` |
 
 ### Authenticate using Microsoft Entra ID
 
@@ -77,10 +87,10 @@ Authenticating with Microsoft Entra ID is supported with Azure Database for Post
 
 | Field | Required | Details | Example |
 |--------------------|:--------:|---------|---------|
-| `timeoutInSeconds` | N | Timeout, in seconds, for all database operations. Defaults to `20` | `30`
 | `tableName` | N | Name of the table where the data is stored. Defaults to `state`. Can optionally have the schema name as prefix, such as `public.state` | `"state"`, `"public.state"`
 | `metadataTableName` | N | Name of the table Dapr uses to store a few metadata properties. Defaults to `dapr_metadata`. Can optionally have the schema name as prefix, such as `public.dapr_metadata` | `"dapr_metadata"`, `"public.dapr_metadata"`
-| `cleanupIntervalInSeconds` | N | Interval, in seconds, to clean up rows with an expired TTL. Default: `3600` (i.e. 1 hour). Setting this to values <=0 disables the periodic cleanup. | `1800`, `-1`
+| `timeout` | N | Timeout for operations on the database, as a [Go duration](https://pkg.go.dev/time#ParseDuration). Integers are interpreted as number of seconds. Defaults to `20s` | `"30s"`, `30` |
+| `cleanupInterval` | N | Interval, as a Go duration or number of seconds, to clean up rows with an expired TTL. Default: `1h` (1 hour). Setting this to values <=0 disables the periodic cleanup. | `"30m"`, `1800`, `-1`
 | `maxConns` | N | Maximum number of connections pooled by this component. Set to 0 or lower to use the default value, which is the greater of 4 or the number of CPUs. | `"4"`
 | `connectionMaxIdleTime` | N | Max idle time before unused connections are automatically closed in the connection pool. By default, there's no value and this is left to the database driver to choose. | `"5m"`
 | `queryExecMode` | N | Controls the default mode for executing queries. By default Dapr uses the extended protocol and automatically prepares and caches prepared statements. However, this may be incompatible with proxies such as PGBouncer. In this case it may be preferrable to use `exec` or `simple_protocol`. | `"simple_protocol"`
@@ -100,8 +110,8 @@ Authenticating with Microsoft Entra ID is supported with Azure Database for Post
 
      > This example does not describe a production configuration because it sets the password in plain text and the user name is left as the PostgreSQL default of "postgres".
 
-2. Create a database for state data.
-Either the default "postgres" database can be used, or create a new database for storing state data.
+2. Create a database for state data.  
+    Either the default "postgres" database can be used, or create a new database for storing state data.
 
     To create a new database in PostgreSQL, run the following SQL command:
 
@@ -121,10 +131,10 @@ This state store supports [Time-To-Live (TTL)]({{< ref state-store-ttl.md >}}) f
 
 Because PostgreSQL doesn't have built-in support for TTLs, this is implemented in Dapr by adding a column in the state table indicating when the data is to be considered "expired". Records that are "expired" are not returned to the caller, even if they're still physically stored in the database. A background "garbage collector" periodically scans the state table for expired rows and deletes them.
 
-The interval at which the deletion of expired records happens is set with the `cleanupIntervalInSeconds` metadata property, which defaults to 3600 seconds (that is, 1 hour). 
+The interval at which the deletion of expired records happens is set with the `cleanupInterval` metadata property, which defaults to 3600 seconds (that is, 1 hour).
 
-- Longer intervals require less frequent scans for expired rows, but can require storing expired records for longer, potentially requiring more storage space. If you plan to store many records in your state table, with short TTLs, consider setting `cleanupIntervalInSeconds` to a smaller value, for example `300` (300 seconds, or 5 minutes).
-- If you do not plan to use TTLs with Dapr and the PostgreSQL state store, you should consider setting `cleanupIntervalInSeconds` to a value <= 0 (e.g. `0` or `-1`) to disable the periodic cleanup and reduce the load on the database.
+- Longer intervals require less frequent scans for expired rows, but can require storing expired records for longer, potentially requiring more storage space. If you plan to store many records in your state table, with short TTLs, consider setting `cleanupInterval` to a smaller value, for example `5m` (5 minutes).
+- If you do not plan to use TTLs with Dapr and the PostgreSQL state store, you should consider setting `cleanupInterval` to a value <= 0 (e.g. `0` or `-1`) to disable the periodic cleanup and reduce the load on the database.
 
 The column in the state table where the expiration date for records is stored in, `expiredate`, **does not have an index by default**, so each periodic cleanup must perform a full-table scan. If you have a table with a very large number of records, and only some of them use a TTL, you may find it useful to create an index on that column. Assuming that your state table name is `state` (the default), you can use this query:
 
--- a/daprdocs/content/en/reference/components-reference/supported-state-stores/setup-postgresql-v1.md
+++ b/daprdocs/content/en/reference/components-reference/supported-state-stores/setup-postgresql-v1.md
@@ -131,10 +131,10 @@ This state store supports [Time-To-Live (TTL)]({{< ref state-store-ttl.md >}}) f
 
 Because PostgreSQL doesn't have built-in support for TTLs, this is implemented in Dapr by adding a column in the state table indicating when the data is to be considered "expired". Records that are "expired" are not returned to the caller, even if they're still physically stored in the database. A background "garbage collector" periodically scans the state table for expired rows and deletes them.
 
-The interval at which the deletion of expired records happens is set with the `cleanupInterval` metadata property, which defaults to 3600 seconds (that is, 1 hour).
+You can set the deletion interval of expired records with the `cleanupInterval` metadata property, which defaults to 3600 seconds (that is, 1 hour).
 
-- Longer intervals require less frequent scans for expired rows, but can require storing expired records for longer, potentially requiring more storage space. If you plan to store many records in your state table, with short TTLs, consider setting `cleanupInterval` to a smaller value, for example `5m` (5 minutes).
-- If you do not plan to use TTLs with Dapr and the PostgreSQL state store, you should consider setting `cleanupInterval` to a value <= 0 (e.g. `0` or `-1`) to disable the periodic cleanup and reduce the load on the database.
+- Longer intervals require less frequent scans for expired rows, but can require storing expired records for longer, potentially requiring more storage space. If you plan to store many records in your state table, with short TTLs, consider setting `cleanupInterval` to a smaller value; for example, `5m` (5 minutes).
+- If you do not plan to use TTLs with Dapr and the PostgreSQL state store, you should consider setting `cleanupInterval` to a value <= 0 (for example, `0` or `-1`) to disable the periodic cleanup and reduce the load on the database.
 
 The column in the state table where the expiration date for records is stored in, `expiredate`, **does not have an index by default**, so each periodic cleanup must perform a full-table scan. If you have a table with a very large number of records, and only some of them use a TTL, you may find it useful to create an index on that column. Assuming that your state table name is `state` (the default), you can use this query:
 
--- a/daprdocs/content/en/reference/components-reference/supported-state-stores/setup-postgresql-v2.md
+++ b/daprdocs/content/en/reference/components-reference/supported-state-stores/setup-postgresql-v2.md
@@ -9,7 +9,7 @@ aliases:
 ---
 
 {{% alert title="Note" color="primary" %}}
-This is the v2 of the PostgreSQL state store component, which contains some improvements to performance and reliability.
+This is the v2 of the PostgreSQL state store component, which contains some improvements to performance and reliability. New applications are encouraged to use v2.
 
 The PostgreSQL v2 state store component is not compatible with the [v1 component]({{< ref setup-postgresql-v1.md >}}), and data cannot be migrated between the two components. The v2 component does not offer support for state store query APIs.
 
--- a/daprdocs/content/en/reference/components-reference/supported-state-stores/setup-postgresql-v2.md
+++ b/daprdocs/content/en/reference/components-reference/supported-state-stores/setup-postgresql-v2.md
@@ -1,0 +1,165 @@
+---
+type: docs
+title: "PostgreSQL"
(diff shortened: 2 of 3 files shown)


daprdocs/data/components/state_stores/generic.yaml
```diff
@@ -141,8 +141,8 @@
     etag: true
     ttl: true
     query: false
-- component: PostgreSQL
-  link: setup-postgresql
+- component: PostgreSQL v1
+  link: setup-postgresql-v1
```
ItalyPaleAle, line 145: Note that the old link is still available as alias

daprdocs/content/en/reference/components-reference/supported-state-stores/setup-postgresql-v1.md
```diff
@@ -21,8 +31,8 @@ spec:
     # Connection string
     - name: connectionString
       value: "<CONNECTION STRING>"
-    # Timeout for database operations, in seconds (optional)
-    #- name: timeoutInSeconds
+    # Timeout for database operations, as a Go duration or number of seconds (optional)
+    #- name: timeout
```
msfussell, line 35: Presume this is not a breaking change?

daprdocs/content/en/reference/components-reference/supported-state-stores/setup-postgresql-v1.md
```diff
@@ -31,8 +41,8 @@ spec:
     #- name: metadataTableName
     #  value: "dapr_metadata"
     # Cleanup interval in seconds, to remove expired rows (optional)
-    #- name: cleanupIntervalInSeconds
-    #  value: 3600
+    #- name: cleanupInterval
```
msfussell, line 44: Presume this is not a breaking change?

daprdocs/content/en/reference/components-reference/supported-state-stores/setup-postgresql-v1.md
```diff
@@ -31,8 +41,8 @@ spec:
...
+    #  value: "1h"
```
msfussell, line 45: Would be best to include an example in seconds also.

daprdocs/content/en/reference/components-reference/supported-state-stores/setup-postgresql-v1.md
```diff
@@ -77,10 +87,10 @@ Authenticating with Microsoft Entra ID is supported with Azure Database for Post
 
 | Field | Required | Details | Example |
 |--------------------|:--------:|---------|---------|
-| `timeoutInSeconds` | N | Timeout, in seconds, for all database operations. Defaults to `20` | `30`
 | `tableName` | N | Name of the table where the data is stored. Defaults to `state`. Can optionally have the schema name as prefix, such as `public.state` | `"state"`, `"public.state"`
 | `metadataTableName` | N | Name of the table Dapr uses to store a few metadata properties. Defaults to `dapr_metadata`. Can optionally have the schema name as prefix, such as `public.dapr_metadata` | `"dapr_metadata"`, `"public.dapr_metadata"`
-| `cleanupIntervalInSeconds` | N | Interval, in seconds, to clean up rows with an expired TTL. Default: `3600` (i.e. 1 hour). Setting this to values <=0 disables the periodic cleanup. | `1800`, `-1`
+| `timeout` | N | Timeout for operations on the database, as a [Go duration](https://pkg.go.dev/time#ParseDuration). Integers are interpreted as number of seconds. Defaults to `20s` | `"30s"`, `30` |
```
msfussell, line 92: Presume this is not a breaking change?

daprdocs/content/en/reference/components-reference/supported-state-stores/setup-postgresql-v1.md
```diff
@@ -77,10 +87,10 @@ Authenticating with Microsoft Entra ID is supported with Azure Database for Post
...
+| `cleanupInterval` | N | Interval, as a Go duration or number of seconds, to clean up rows with an expired TTL. Default: `1h` (1 hour). Setting this to values <=0 disables the periodic cleanup. | `"30m"`, `1800`, `-1`
```
msfussell, line 93: Presume this is not a breaking change?

daprdocs/content/en/reference/components-reference/supported-state-stores/setup-postgresql-v2.md
```diff
@@ -0,0 +1,165 @@
...
+title: "PostgreSQL"
+linkTitle: "PostgreSQL"
+description: Detailed information on the PostgreSQL state store component
+aliases:
+  - "/operations/components/setup-state-store/supported-state-stores/setup-postgresql-v2/"
+  - "/operations/components/setup-state-store/supported-state-stores/setup-postgres-v2/"
+---
+
+{{% alert title="Note" color="primary" %}}
+This is the v2 of the PostgreSQL state store component, which contains some improvements to performance and reliability.
```
msfussell: There are other reasons that you state here https://github.com/dapr/docs/pull/3996/files including 
- CockroachDB, YugabyteDB - Should users use this as the preferred component for these over https://docs.dapr.io/reference/components-reference/supported-state-stores/setup-cockroachdb/. If so, would be worth adding something into these component specs to say this

- Is this the preferred version to use over the V1 component? If so, suggest saying this.

daprdocs/content/en/reference/components-reference/supported-state-stores/setup-postgresql-v2.md
```diff
@@ -0,0 +1,165 @@
...
+    #  value: 20
+    # Prefix for the table where the data is stored (optional)
+    #- name: tablePrefix
+    #  value: ""
+    # Name of the table where to store metadata used by Dapr (optional)
+    #- name: metadataTableName
+    #  value: "dapr_metadata"
+    # Cleanup interval in seconds, to remove expired rows (optional)
+    #- name: cleanupInterval
+    #  value: "1h"
```
msfussell, line 45: Add a seconds example.

msfussell (commented):
A few comments.

daprdocs/content/en/reference/components-reference/supported-state-stores/setup-postgresql-v1.md
ItalyPaleAle: Correct, the old value is aliased. The new one is more flexible as it allows seconds or a Go duration.

daprdocs/content/en/reference/components-reference/supported-state-stores/setup-postgresql-v1.md
ItalyPaleAle: Ditto

daprdocs/content/en/reference/components-reference/supported-state-stores/setup-postgresql-v1.md
ItalyPaleAle: It should be included below in the table, or did i miss that?

daprdocs/content/en/reference/components-reference/supported-state-stores/setup-postgresql-v1.md
ItalyPaleAle: Ditto

daprdocs/content/en/reference/components-reference/supported-state-stores/setup-postgresql-v1.md
ItalyPaleAle: Ditto

daprdocs/content/en/reference/components-reference/supported-state-stores/setup-postgresql-v2.md
ItalyPaleAle: CockroachDB will work but not sure if it should have its own component?

As for YugabyteDB, our tests show that it requires a bit of work at this stage. Specifically we need to make users enable the crypto extension for using the UUID generator function

daprdocs/content/en/reference/components-reference/supported-state-stores/setup-postgresql-v2.md
ItalyPaleAle: Should be in the table below

daprdocs/content/en/reference/components-reference/supported-state-stores/setup-postgresql-v2.md
```diff
@@ -0,0 +1,165 @@
...
+
+### Authenticate using Microsoft Entra ID
+
+Authenticating with Microsoft Entra ID is supported with Azure Database for PostgreSQL. All authentication methods supported by Dapr can be used, including client credentials ("service principal") and Managed Identity.
+
+| Field  | Required | Details | Example |
+|--------|:--------:|---------|---------|
+| `useAzureAD` | Y | Must be set to `true` to enable the component to retrieve access tokens from Microsoft Entra ID. | `"true"` |
+| `connectionString` | Y | The connection string for the PostgreSQL database.<br>This must contain the user, which corresponds to the name of the user created inside PostgreSQL that maps to the Microsoft Entra ID identity; this is often the name of the corresponding principal (e.g. the name of the Microsoft Entra ID application). This connection string should not contain any password.  | `"host=mydb.postgres.database.azure.com user=myapplication port=5432 database=my_db sslmode=require"` |
```
hhunter-ms:
```suggestion
| `connectionString` | Y | The connection string for the PostgreSQL database.<br>This must contain the user, which corresponds to the name of the user created inside PostgreSQL that maps to the Microsoft Entra ID identity. This is often the name of the corresponding principal (for example, the name of the Microsoft Entra ID application). This connection string should not contain any password.  | `"host=mydb.postgres.database.azure.com user=myapplication port=5432 database=my_db sslmode=require"` |
```

daprdocs/content/en/reference/components-reference/supported-state-stores/setup-postgresql-v2.md
```diff
@@ -0,0 +1,165 @@
...
+
+Because PostgreSQL doesn't have built-in support for TTLs, this is implemented in Dapr by adding a column in the state table indicating when the data is to be considered "expired". Records that are "expired" are not returned to the caller, even if they're still physically stored in the database. A background "garbage collector" periodically scans the state table for expired rows and deletes them.
+
+The interval at which the deletion of expired records happens is set with the `cleanupInterval` metadata property, which defaults to 3600 seconds (that is, 1 hour).
+
+- Longer intervals require less frequent scans for expired rows, but can require storing expired records for longer, potentially requiring more storage space. If you plan to store many records in your state table, with short TTLs, consider setting `cleanupInterval` to a smaller value, for example `5m` (5 minutes).
+- If you do not plan to use TTLs with Dapr and the PostgreSQL state store, you should consider setting `cleanupInterval` to a value <= 0 (e.g. `0` or `-1`) to disable the periodic cleanup and reduce the load on the database.
```
hhunter-ms:
```suggestion
- If you do not plan to use TTLs with Dapr and the PostgreSQL state store, you should consider setting `cleanupInterval` to a value <= 0 (for example, `0` or `-1`) to disable the periodic cleanup and reduce the load on the database.
```

daprdocs/content/en/reference/components-reference/supported-state-stores/setup-postgresql-v1.md
```diff
@@ -121,10 +131,10 @@ This state store supports [Time-To-Live (TTL)]({{< ref state-store-ttl.md >}}) f
...
 
-- Longer intervals require less frequent scans for expired rows, but can require storing expired records for longer, potentially requiring more storage space. If you plan to store many records in your state table, with short TTLs, consider setting `cleanupIntervalInSeconds` to a smaller value, for example `300` (300 seconds, or 5 minutes).
-- If you do not plan to use TTLs with Dapr and the PostgreSQL state store, you should consider setting `cleanupIntervalInSeconds` to a value <= 0 (e.g. `0` or `-1`) to disable the periodic cleanup and reduce the load on the database.
+- Longer intervals require less frequent scans for expired rows, but can require storing expired records for longer, potentially requiring more storage space. If you plan to store many records in your state table, with short TTLs, consider setting `cleanupInterval` to a smaller value, for example `5m` (5 minutes).
+- If you do not plan to use TTLs with Dapr and the PostgreSQL state store, you should consider setting `cleanupInterval` to a value <= 0 (e.g. `0` or `-1`) to disable the periodic cleanup and reduce the load on the database.
```
hhunter-ms:
```suggestion
- If you do not plan to use TTLs with Dapr and the PostgreSQL state store, you should consider setting `cleanupInterval` to a value <= 0 (for example, `0` or `-1`) to disable the periodic cleanup and reduce the load on the database.
```

daprdocs/content/en/reference/components-reference/supported-state-stores/setup-postgresql-v1.md
```diff
@@ -121,10 +131,10 @@ This state store supports [Time-To-Live (TTL)]({{< ref state-store-ttl.md >}}) f
...
+The interval at which the deletion of expired records happens is set with the `cleanupInterval` metadata property, which defaults to 3600 seconds (that is, 1 hour).
 
-- Longer intervals require less frequent scans for expired rows, but can require storing expired records for longer, potentially requiring more storage space. If you plan to store many records in your state table, with short TTLs, consider setting `cleanupIntervalInSeconds` to a smaller value, for example `300` (300 seconds, or 5 minutes).
-- If you do not plan to use TTLs with Dapr and the PostgreSQL state store, you should consider setting `cleanupIntervalInSeconds` to a value <= 0 (e.g. `0` or `-1`) to disable the periodic cleanup and reduce the load on the database.
+- Longer intervals require less frequent scans for expired rows, but can require storing expired records for longer, potentially requiring more storage space. If you plan to store many records in your state table, with short TTLs, consider setting `cleanupInterval` to a smaller value, for example `5m` (5 minutes).
```
hhunter-ms:
```suggestion
- Longer intervals require less frequent scans for expired rows, but can require storing expired records for longer, potentially requiring more storage space. If you plan to store many records in your state table, with short TTLs, consider setting `cleanupInterval` to a smaller value; for example, `5m` (5 minutes).
```

daprdocs/content/en/reference/components-reference/supported-state-stores/setup-postgresql-v2.md
```diff
@@ -0,0 +1,165 @@
...
+  
+{{% /codetab %}}
+
+{{% /tabs %}}
+
+## Advanced
+
+### Differences between v1 and v2
+
+The PostgreSQL state store v2 was introduced in Dapr 1.13. The pre-existing v1 remains available and is not deprecated.
```
hhunter-ms:
```suggestion
The PostgreSQL state store v2 was introduced in Dapr 1.13. The [pre-existing v1]({{< ref setup-postgresql-v1.md >}}) remains available and is not deprecated.
```

daprdocs/content/en/reference/components-reference/supported-state-stores/setup-postgresql-v2.md
```diff
@@ -0,0 +1,165 @@
...
+
+In the v2 component, the table schema has been changed significantly, with the goal of increasing performance and reliability. Most notably, the value stored by Dapr is now of type _BYTEA_, which allows faster queries and in some cases it's more space-efficient than the previously-used _JSONB_ column.
```
hhunter-ms:
```suggestion
In the v2 component, the table schema has been changed significantly, with the goal of increasing performance and reliability. Most notably, the value stored by Dapr is now of type _BYTEA_, which allows faster queries and, in some cases, is more space-efficient than the previously-used _JSONB_ column.  
```

daprdocs/content/en/reference/components-reference/supported-state-stores/setup-postgresql-v2.md
```diff
@@ -0,0 +1,165 @@
...
+{{% /tabs %}}
+
+## Advanced
+
+### Differences between v1 and v2
+
+The PostgreSQL state store v2 was introduced in Dapr 1.13. The pre-existing v1 remains available and is not deprecated.
+
+In the v2 component, the table schema has been changed significantly, with the goal of increasing performance and reliability. Most notably, the value stored by Dapr is now of type _BYTEA_, which allows faster queries and in some cases it's more space-efficient than the previously-used _JSONB_ column.  
+On the downside, because of the change above, the v2 component does not support the Dapr state store query APIs.
```
hhunter-ms:
```suggestion
However, due to this change, the v2 component does not support the [Dapr state store query APIs]({{< ref howto-state-query-api.md >}}).
```

daprdocs/content/en/reference/components-reference/supported-state-stores/setup-postgresql-v2.md
```diff
@@ -0,0 +1,165 @@
...
+
+Additionally, etags are now random UUIDs, which ensures better compatibility with other PostgreSQL-compatible databases such as CockroachDB.
```
hhunter-ms:
```suggestion
Also, in the v2 component, ETags are now random UUIDs, which ensures better compatibility with other PostgreSQL-compatible databases, such as CockroachDB.
```

daprdocs/content/en/reference/components-reference/supported-state-stores/setup-postgresql-v2.md
```diff
@@ -0,0 +1,165 @@
...
+
+Because of the changes, it is not possible for v1 and v2 components to read or write data from the same table. At this stage, it is also not possible to migrate data between the two versions of the component.
```
hhunter-ms:
```suggestion
Because of these changes, v1 and v2 components are not able to read or write data from the same table. At this stage, it's also impossible to migrate data between the two versions of the component.
```

daprdocs/content/en/reference/components-reference/supported-state-stores/setup-postgresql-v2.md
```diff
@@ -0,0 +1,165 @@
...
+
+### Displaying the data in human-readable format
+
+The PostgreSQL v2 component stores the state's value in the `value` column, which is of type _BYTEA_. Most PostgreSQL tools, including pgAdmin, consider the value as binary and do not display it in human-readable form by default.
+
+If you want to inspect the value in the state store, and you know it's not binary (for example, JSON data), you can use a query similar to this to have the value displayed in human-readable form:
```
hhunter-ms:
```suggestion
If you want to inspect the value in the state store, and you know it's not binary (for example, JSON data), you can have the value displayed in human-readable form using a query like the following:
```

daprdocs/content/en/reference/components-reference/supported-state-stores/setup-postgresql-v2.md
```diff
@@ -0,0 +1,165 @@
...
+SELECT *, convert_from(value, 'utf-8') FROM state;
+```
+
+### TTLs and cleanups
+
+This state store supports [Time-To-Live (TTL)]({{< ref state-store-ttl.md >}}) for records stored with Dapr. When storing data using Dapr, you can set the `ttlInSeconds` metadata property to indicate after how many seconds the data should be considered "expired".
+
+Because PostgreSQL doesn't have built-in support for TTLs, this is implemented in Dapr by adding a column in the state table indicating when the data is to be considered "expired". Records that are "expired" are not returned to the caller, even if they're still physically stored in the database. A background "garbage collector" periodically scans the state table for expired rows and deletes them.
+
+The interval at which the deletion of expired records happens is set with the `cleanupInterval` metadata property, which defaults to 3600 seconds (that is, 1 hour).
```
hhunter-ms:
```suggestion
You can set the deletion interval of expired records with the `cleanupInterval` metadata property, which defaults to 3600 seconds (that is, 1 hour).
```

daprdocs/content/en/reference/components-reference/supported-state-stores/setup-postgresql-v2.md
```diff
@@ -0,0 +1,165 @@
...
+
+- Longer intervals require less frequent scans for expired rows, but can require storing expired records for longer, potentially requiring more storage space. If you plan to store many records in your state table, with short TTLs, consider setting `cleanupInterval` to a smaller value, for example `5m` (5 minutes).
```
hhunter-ms:
```suggestion
- Longer intervals require less frequent scans for expired rows, but can require storing expired records for longer, potentially requiring more storage space. If you plan to store many records in your state table, with short TTLs, consider setting `cleanupInterval` to a smaller value; for example, `5m` (5 minutes).
```

daprdocs/content/en/reference/components-reference/supported-state-stores/setup-postgresql-v2.md
```diff
@@ -0,0 +1,165 @@
...
+| `timeout` | N | Timeout for operations on the database, as a [Go duration](https://pkg.go.dev/time#ParseDuration). Integers are interpreted as number of seconds. Defaults to `20s` | `"30s"`, `30` |
+| `cleanupInterval` | N | Interval, as a Go duration or number of seconds, to clean up rows with an expired TTL. Default: `1h` (1 hour). Setting this to values <=0 disables the periodic cleanup. | `"30m"`, `1800`, `-1` |
+| `maxConns` | N | Maximum number of connections pooled by this component. Set to 0 or lower to use the default value, which is the greater of 4 or the number of CPUs. | `"4"` |
+| `connectionMaxIdleTime` | N | Max idle time before unused connections are automatically closed in the connection pool. By default, there's no value and this is left to the database driver to choose. | `"5m"` |
+| `queryExecMode` | N | Controls the default mode for executing queries. By default Dapr uses the extended protocol and automatically prepares and caches prepared statements. However, this may be incompatible with proxies such as PGBouncer. In this case it may be preferrable to use `exec` or `simple_protocol`. | `"simple_protocol"` |
```
hhunter-ms:
```suggestion
| `queryExecMode` | N | Controls the default mode for executing queries. By default Dapr uses the extended protocol and automatically prepares and caches prepared statements. However, this may be incompatible with proxies such as PGBouncer. In this case, it may be preferrable to use `exec` or `simple_protocol`. | `"simple_protocol"` |
```

daprdocs/content/en/reference/components-reference/supported-state-stores/setup-postgresql-v1.md
```diff
@@ -121,10 +131,10 @@ This state store supports [Time-To-Live (TTL)]({{< ref state-store-ttl.md >}}) f
 
 Because PostgreSQL doesn't have built-in support for TTLs, this is implemented in Dapr by adding a column in the state table indicating when the data is to be considered "expired". Records that are "expired" are not returned to the caller, even if they're still physically stored in the database. A background "garbage collector" periodically scans the state table for expired rows and deletes them.
 
-The interval at which the deletion of expired records happens is set with the `cleanupIntervalInSeconds` metadata property, which defaults to 3600 seconds (that is, 1 hour). 
+The interval at which the deletion of expired records happens is set with the `cleanupInterval` metadata property, which defaults to 3600 seconds (that is, 1 hour).
```
hhunter-ms:
```suggestion
You can set the deletion interval of expired records with the `cleanupInterval` metadata property, which defaults to 3600 seconds (that is, 1 hour).
```

daprdocs/content/en/reference/components-reference/supported-state-stores/setup-postgresql-v1.md
```diff
@@ -100,8 +110,8 @@ Authenticating with Microsoft Entra ID is supported with Azure Database for Post
 
      > This example does not describe a production configuration because it sets the password in plain text and the user name is left as the PostgreSQL default of "postgres".
 
-2. Create a database for state data.
-Either the default "postgres" database can be used, or create a new database for storing state data.
+2. Create a database for state data.
```
hhunter-ms:
```suggestion
1. Create a database for state data.  
```

hhunter-ms (changes requested):
tiny review

daprdocs/content/en/reference/components-reference/supported-state-stores/setup-postgresql-v1.md
ItalyPaleAle: There's a #1 above

daprdocs/content/en/reference/components-reference/supported-state-stores/setup-postgresql-v2.md
msfussell: @ItalyPaleAle. A suggestion and comment

- "This is the v2 of the PostgreSQL state store component, which contains some improvements to performance and reliability. It is recommended to use the V2 component over the V1."

- Regarding "CockroachDB will work but not sure if it should have its own component?" - Should users ever use this component for CockroachDB then, or stay with the specific component https://docs.dapr.io/reference/components-reference/supported-state-stores/setup-cockroachdb/ ? Not sure what you mean by "CockroachDB will work but not sure if it should have its own component?" since it already has its own component.

daprdocs/content/en/reference/components-reference/supported-state-stores/setup-postgresql-v1.md
hhunter-ms: We can have it either way, honestly - in Markdown, only the first number in a numbered list matters. I simply changed this one bc I typically use:

```
1. text
1. text
1. text
```

bc it will render as:

```
1. text
2. text
3. text
```

and I don't have to think about it haha...

source: https://www.markdownguide.org/basic-syntax/#ordered-lists

hhunter-ms (approved):

daprdocs/content/en/reference/components-reference/supported-state-stores/setup-postgresql-v2.md
ItalyPaleAle: CockroachDB currently has its own documentation, not sure if we should have a v2 of that too

ItalyPaleAle:
```suggestion
This is the v2 of the PostgreSQL state store component, which contains some improvements to performance and reliability. New applications are encouraged to use v2.
```

msfussell: Thanks.

PR #4051, opened 2024-02-28
Add pre-reqs and section in troubleshooting guide for `dapr init`

### LINKED ISSUE #3540: Docker Desktop requires socket setting before using Dapr CLI
We need doc to mention that CLI needs this setting in DockerDesktop: "Allow the default Docker socket to be used"

Reproduced this on M1 Mac.

image

Without this, users get this error:
```
$ dapr init
⌛  Making the jump to hyperspace...
❌  could not connect to docker. docker may not be installed or running
```

Even though `docker ps` works just fine.

Mention the setting as a requirement for using Dapr CLI - OR mention in troubleshooting guide

https://docs.dapr.io/getting-started/install-dapr-cli/
OR here: https://docs.dapr.io/operations/troubleshooting/common_issues/

None - it seems to be a situation I ran into in Docker Desktop itself.

**Additional context**
DOES NOT seem to be a breaking change in Docker Desktop.

hhunter-ms:
## Description

Add prereq note before installing CLI to make sure default Docker socket is enabled.
Add link to troubleshooting guide after `dapr init` step in case of error.
Add section in troubleshooting guide regarding error received about Docker not running/installed after `dapr init`

## Issue reference

PR will close: #3540


msfussell (approved):
LGTM

PR #4082, opened 2024-03-15
add translation in zh-Hans for v1.13

newbe36524:
## Description

update zh-Hans translation to the latest content according to the 1.13 branch

all content has been translated, except "operation" & "reference" are not done since there is a huge works to do with those sections.


hhunter-ms (approved):
lgtm

newbe36524:
@hhunter-ms If you can, I think you can just merge the code. Because I don't have the corresponding permissions to operate. In the past, after we provided the translation results, the current administrator would merge them. Thanks

PR #4083, opened 2024-03-18
Fix the copy/paste typo

### LINKED ISSUE #4081: copy-paste error in golang example
**URL of the docs page**
https://docs.dapr.io/developing-applications/building-blocks/workflow/workflow-patterns/

**How is it currently worded?**
```
	if err := ctx.CallActivity(Step1, workflow.ActivityInput(input)).Await(&result1); err != nil {
		return nil, err
	}
	var result2 int
	if err := ctx.CallActivity(Step1, workflow.ActivityInput(input)).Await(&result2); err != nil {
		return nil, err
	}
	var result3 int
	if err := ctx.CallActivity(Step1, workflow.ActivityInput(input)).Await(&result3); err != nil {
		return nil, err
	}
```

**How should it be worded?**
The `Step2` and `Step3` should be passed into the last two calls to `CallActivity(...)`.  Right now it's passing `Step1` to all three function invocations.  I think it's just a copy/paste mistake.
```
	if err := ctx.CallActivity(Step1, workflow.ActivityInput(input)).Await(&result1); err != nil {
		return nil, err
	}
	var result2 int
	if err := ctx.CallActivity(Step2, workflow.ActivityInput(input)).Await(&result2); err != nil {
		return nil, err
	}
	var result3 int
	if err := ctx.CallActivity(Step3, workflow.ActivityInput(input)).Await(&result3); err != nil {
		return nil, err
	}
```

hhunter-ms:
## Description

Small edit to Go code snippet in Workflow Patterns doc

## Issue reference

PR will close: #4081


marcduiker (approved):
LGTM

PR #4128, opened 2024-05-03
Upmerge 1.13 --> 1.14

hhunter-ms:


.github/workflows/website-root.yml
```diff
@@ -0,0 +1,110 @@
+name: Azure Static Web App Root
```
msfussell: Actually I remember now. .github/workflows/website-root.yml  should only be for the current repo v1.13.  If you look at v1.12 repo is it not there. So you should remove this file from the upmerge.

msfussell (approved):
LGTM

PR #4151, opened 2024-05-18
Adding a compression cost optimization section to the Cosmos DB docs

### LINKED ISSUE #4047: CosmosDB docs should contain a recommendation on how to save costs when using actor state management to handle large documents
_Feel free to throw this issue out the window if you believe it doesn't belong in the Dapr docs, but I thought it might be nice to have :)_

The Dapr [CosmosDB state management documentation](https://docs.dapr.io/reference/components-reference/supported-state-stores/setup-azure-cosmosdb/) should maybe mention that CosmosDB's billing profile is a consumption-based billing profile. This means that when using Dapr to read and write state, more specifically Actor state which is raw, formatted JSON, if the actor's state object grows in size, it will consume a fair chunk of money to run. This grows linearly with document size and concurrency. One way to combat this growth in cost is to compress the document using something like GZip before saving state and decompressing when getting state.

Something simple like a new heading on savings costs. Could look as simple as:

> # Saving costs when using actors to store large documents
> 
> Cosmos DB uses a consumption-based billing model. This means that if you have an actor state object that is large in size, you will pay for every KB you write on every actor call and for every get each time an actor is hydrated. For example, if you have an actor state object that is 75KB in size when the actor is hydrated, you will use 75 RU/s to read that object out of state. If you then modify the state object and it grows to 100KB, you will use 100 RU/s to write that object to Cosmos DB. Totalling 175 RU/s for the I/O operation. Let's say your actors are concurrently handling 1000 requests per second, you will need at least 175,000 RU/s to meet that load. 
> 
> This is largely because Dapr actor state is saved as raw, formatted JSON in the state store. Furthermore, Dapr interacts with state as key-value pairs and as such, there is no definite need to store the actor state as raw JSON. One way to save costs is to compress the state object before saving. This can be done with any compression algorithm but ideally pick one that favours speed over outright compression optimisation. 
>
> Done correctly, this can reduce costs by 95% or more.

It could also be worth linking to this header where it mentions:

> If you are using the Dapr SDKs (for example the [.NET SDK](https://github.com/dapr/dotnet-sdk)), the SDK automatically serializes your data to JSON.

Under the [Azure Cosmos DB (SQL API)](https://docs.dapr.io/reference/components-reference/supported-state-stores/setup-azure-cosmosdb) page at the bottom below the [Optimizing Cosmos DB for bulk operation write performance](https://docs.dapr.io/reference/components-reference/supported-state-stores/setup-azure-cosmosdb#optimizing-cosmos-db-for-bulk-operation-write-performance) header.

**Additional context**
This is inspired by some good work one of my colleagues has recently completed and published an article about: [Maximizing Efficiency in Scalable Systems: Part I — Reducing costs of consumption-based state stores by up to 97% when using Dapr actors at scale](https://spharah.medium.com/maximizing-efficiency-in-scalable-systems-part-i-reducing-costs-of-consumption-based-state-d114ffc17f80)

KrylixZA:
## Description

This PR adds further optimisations to the Cosmos DB state management, focusing on cost savings. The section breaks down how Cosmos DB RU/s are calculated and how developers can use compression algorithms to reduce the overall size of the object they are reading and writing to state and subsequently reduce the number of RU/s used. This is inspired by the work of my colleague who reduced our RU usage by 97% which he wrote about here: [Maximizing Efficiency in Scalable Systems: Part I — Reducing costs of consumption-based state stores by up to 97% when using Dapr actors at scale](https://spharah.medium.com/maximizing-efficiency-in-scalable-systems-part-i-reducing-costs-of-consumption-based-state-d114ffc17f80)

## Issue reference

Closes #4047


hhunter-ms (approved):

msfussell (approved):
LGTM. Thanks for the contribution @KrylixZA !

PR #4286, opened 2024-07-31
Fix outbox projection http json

yaron2:
Fixes JSON for HTTP sample.

### DIFF
--- a/daprdocs/content/en/developing-applications/building-blocks/state-management/howto-outbox.md
+++ b/daprdocs/content/en/developing-applications/building-blocks/state-management/howto-outbox.md
@@ -110,11 +110,11 @@ spec:
 
 ### Shape the outbox pattern message
 
-You can override the outbox pattern message published to the pub/sub broker by setting a different message. This is done via a projected transaction payload, called `outboxProjections`, which is ignored when the state is written and is used as the outbox pattern message published to the user topic.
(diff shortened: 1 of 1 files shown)


daprdocs/content/en/developing-applications/building-blocks/state-management/howto-outbox.md
```diff
@@ -110,11 +110,11 @@ spec:
 
 ### Shape the outbox pattern message
 
-You can override the outbox pattern message published to the pub/sub broker by setting a different message. This is done via a projected transaction payload, called `outboxProjections`, which is ignored when the state is written and is used as the outbox pattern message published to the user topic.
+You can override the outbox pattern message published to the pub/sub broker by setting another transaction that will not be saved to the database and is explicitly mentioned as a projection. This transaction is added a metadata key named `outbox.projection` with a value set to `true`. When added to the state array saved in a transaction, this payload will be ignored when the state is written and the data will be used as the payload sent to the upstream subscriber.
```
msfussell:
```suggestion
You can override the outbox pattern message published to the pub/sub broker by setting another transaction that is not be saved to the database and is explicitly mentioned as a projection. This transaction is added a metadata key named `outbox.projection` with a value set to `true`. When added to the state array saved in a transaction, this payload is ignored when the state is written and the data is used as the payload sent to the upstream subscriber.
```

msfussell (approved):
LGTM

PR #4292, opened 2024-08-01
[Workflow] Remove text about trace context support

hhunter-ms:
## Description

Workflow doesn't have access to the current trace context.

## Issue reference

PR will close: #4271


msfussell (approved):
LGTM

PR #4324, opened 2024-08-19
[Annotations] Clarify what 'empty' value means

### LINKED ISSUE #4044: Clearly document how to disable placement connection on non-actor services
**Describe the issue**

On the page: [Dapr arguments and annotations for daprd, CLI, and Kubernetes](https://docs.dapr.io/reference/arguments-annotations-overview/), it specifies

> When the annotation is set and the value is empty, the sidecar does not connect to Placement server. This can be used when there are no actors running in the sidecar.

However it is not completely clear that the value needs to be a single space. Simply putting `''` does not work. It needs to be set to `' '` for this to work.

**URL of the docs**
<https://docs.dapr.io/reference/arguments-annotations-overview/>

**Expected content**
The docs should explicitly state that it needs to be set to `' '` in order for it to work and specify this needs to be a single space character.

**Screenshots**
N/A

**Additional context**
This is for those wanting to get rid of the `404 /dapr/config` issues in their telemetry during application startup.

hhunter-ms:
## Description

Clarify the annotations description for placement and scheduler services. "Empty" value means "single space"

## Issue reference

PR will close: #4044


hhunter-ms:
@KrylixZA for your review - thanks for bringing this up!

msfussell (approved):
LGTM

PR #4339, opened 2024-09-04
[Resource specs] Add note about namespacing

### LINKED ISSUE #4117: Make it clear that Application Id Scopes and Namespaces apply to all Dapr resources
It should be made clear that  all resources today can be namespaced and all resources other than configuration can be scoped to a set of application IDs. These are the resources https://docs.dapr.io/reference/resource-specs/ and I suggest that this is pulled into each of the articles for each resource.

Configurations are different from other resources because there's only a 1:n relationship between an app and a configuration (an app can only have one configuration) so the app is scoping them by selecting the configuration explicitly

hhunter-ms:
## Description

Add note to applicable resource specs about namespacing and restricting access.
Make examples consistent.

## Issue reference

PR will close: #4117

### DIFF
--- a/daprdocs/content/en/reference/resource-specs/component-schema.md
+++ b/daprdocs/content/en/reference/resource-specs/component-schema.md
@@ -8,8 +8,10 @@ description: "The basic spec for a Dapr component"
 
 Dapr defines and registers components using a [resource specifications](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/). All components are defined as a resource and can be applied to any hosting environment where Dapr is running, not just Kubernetes.
 
+Typically, components are restricted to a particular [namepsace]({{< ref isolation-concept.md >}}) and restricted access through scopes to any particular set of applications. The namespace is either explicit on the component manifest itself, or set by the API server, which derives the namespace through context with applying to Kubernetes. 
+
 {{% alert title="Note" color="primary" %}}
-Any component can be restricted to a particular [namepsace]({{< ref isolation-concept.md >}}) and restricted access through scopes to any particular set of applications.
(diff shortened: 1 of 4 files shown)


msfussell (approved):
LGTM

daprdocs/content/en/reference/resource-specs/component-schema.md
```diff
@@ -8,27 +8,31 @@ description: "The basic spec for a Dapr component"
 
 Dapr defines and registers components using a [resource specifications](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/). All components are defined as a resource and can be applied to any hosting environment where Dapr is running, not just Kubernetes.
 
+{{% alert title="Note" color="primary" %}}
+Any component can be restricted to a particular [namepsace]({{< ref isolation-concept.md >}}) and restricted access through scopes to any particular set of applications.
+{{% /alert %}}
```
JoshVanL: I think it is worth noting that a Component is _always_ restricted to a particular namespace- there is no concept of a "global" Component which can be ingested by all daprds regardless of namespace. This namespace is either explicit on the Component manifest itself, or set by the API server which derives the namespace through context when applying to Kubernetes. The only exception to this rule is in self-hosted mode whereby daprd will ingest Component (or other) resources when the namespace field is omitted, though the security profile of this is mute as daprd has access to the manifest anyway, unlike in kube.

JoshVanL (changes requested):

daprdocs/content/en/reference/resource-specs/component-schema.md
hhunter-ms: @JoshVanL thanks! I tried to clarify this in the doc, if you could take a look?

JoshVanL: LGTM!

JoshVanL (approved):

PR #4342, opened 2024-09-06
Updates doc with new Scheduler volume size default option.

JoshVanL:
Hold until https://github.com/dapr/dapr/pull/8061


msfussell (approved):
LGTM

JoshVanL:
Default is no longer true after v1.14.4

PR #4403, opened 2024-10-25
Update subscriptions examples

### LINKED ISSUE #4322: Documentation/Tutorials for setting up PubSub component are unclear
**Describe the issue**
- Declarative subscription guide illustrates subscribing to a `orders` topic with route to the subscription as `/checkout`. This is already confusing since the example uses `checkout` entity as being the 'publisher'. Why is the publisher subscribing..? 
- Programmatic subscription guide with Python illustrates subscribing to a `checkout` topic with route to the subscription as `/order`. This is breaking consistency with the first example which the doc states is being replicated in the prog-subscription documentation. For folks like me, who don't know better, this will cause a lot of confusion. 
- In the declarative subscription, why does `scopes` include both `checkout` and `orderprocessing` ..? isn't it supposed to be just `orderprocessing` that is subscribing..? 

**URL of the docs**
https://docs.dapr.io/developing-applications/building-blocks/pubsub/subscription-methods/#declarative-subscriptions
https://docs.dapr.io/developing-applications/building-blocks/pubsub/subscription-methods/#programmatic-subscriptions

**Expected content**

**Screenshots**

**Additional context**

hhunter-ms:
## Description

Update endpoints to correlate with actual examples from quickstarts repo
Checking on scopes question in the issue

## Issue reference

PR will close: #4322


msfussell (approved):
LGTM

PR #4413, opened 2024-10-31
Updates documentation for circuit breaker with possible values and adds examples

elena-kolevska:


hhunter-ms:
@elena-kolevska under `trip`, can you have more than one value? ie: `totalFailures` _and_ `requests`? If so, can you add to the code example below the table to show the syntax for having two values? would it look something like this: `request > 5 && totalFailures > 3`

elena-kolevska:
Great question. I did some tests and `totalFailures > 2 || requests > 5` for example is accepted, but the behaviour is not as you'd expect. I'll move this to draft and look some more into it.

msfussell:
@elena-kolevska  - Any update here?

msfussell (approved):
LGTM

PR #4482, opened 2025-01-06
Update python versions for link validation

hhunter-ms:
Updating python versions for 1.13 - it's the last one of the more recent Dapr releases that needs updated


msfussell (approved):
LGTM

PR #4528, opened 2025-02-11
[Conversation] Add HTTP python and js tabs to the quickstart

hhunter-ms:
Add tabs with instructions for the HTTP samples until SDK is available


msfussell (commented):
Looks good. 

The one fix is that I notice the conversation API needs to be added into the table here https://v1-15.docs.dapr.io/getting-started/quickstarts/ (it is in the index currently)

Also in the same file I think it is time we change this note "
We are actively working on adding to our quickstart library. In the meantime, you can explore Dapr through our [tutorials]"

to say something like this? Thoughts?


"Each release, the quickstart library has new examples added for the APIs and SDKs. You can also explore Dapr through the [tutorials](https://v1-15.docs.dapr.io/getting-started/tutorials/).

msfussell (approved):
LGTM

PR #4662, opened 2025-05-22
doc: gcp bulkGet, copy, rename, move docs

nelson-parente:

- [ ] [Read the contribution guide](https://docs.dapr.io/contributing/docs-contrib/contributing-docs/)

## Description

Docs for the operations added here: https://github.com/dapr/components-contrib/pull/3810
Docs for bulkGet added here: https://github.com/dapr/components-contrib/pull/3811

## Issue reference

https://github.com/dapr/components-contrib/issues/3157
https://github.com/dapr/components-contrib/issues/3655

### DIFF
--- a/daprdocs/content/en/reference/components-reference/supported-bindings/gcpbucket.md
+++ b/daprdocs/content/en/reference/components-reference/supported-bindings/gcpbucket.md
@@ -77,8 +77,13 @@ This component supports **output binding** with the following operations:
 
 - `create` : [Create file](#create-file)
 - `get` : [Get file](#get-file)
+- `bulkGet` : [Bulk get objects](#bulk-get-objects)
 - `delete` : [Delete file](#delete-file)
 - `list`: [List file](#list-files)
+- `copy`: [Copy file](#copy-files)
+- `move`: [Move file](#move-files)
+- `rename`: [Rename file](#rename-files)
+
 
 ### Create file
 
@@ -211,6 +216,72 @@ The metadata parameters are:
 
 The response body contains the value stored in the object.
 
+### Bulk get objects
+
+To perform a bulk get operation that retrieves all bucket files at once, invoke the GCP bucket binding with a `POST` method and the following JSON body:
+
+```json
+{
+  "operation": "bulkGet",
+}
+```
+
+The metadata parameters are:
+
+- `encodeBase64` - (optional) configuration to encode base64 file content before return the content for all files
+
+#### Example
+
+{{< tabs Windows Linux >}}
+
+  {{% codetab %}}
+  ```bash
+  curl -d '{ \"operation\": \"bulkget\"}' http://localhost:<dapr-port>/v1.0/bindings/<binding-name>
+  ```
+  {{% /codetab %}}
+
+  {{% codetab %}}
+  ```bash
+  curl -d '{ "operation": "bulkget"}' \
+        http://localhost:<dapr-port>/v1.0/bindings/<binding-name>
+  ```
+  {{% /codetab %}}
+
+{{< /tabs >}}
+
+#### Response
+
+The response body contains an array of objects, where each object represents a file in the bucket with the following structure:
+
+```json
+[
+  {
+    "name": "file1.txt",
+    "data": "content of file1",
+    "attrs": {
+      "bucket": "mybucket",
+      "name": "file1.txt",
+      "size": 1234,
+      ...
+    }
+  },
+  {
+    "name": "file2.txt",
+    "data": "content of file2",
+    "attrs": {
+      "bucket": "mybucket",
+      "name": "file2.txt",
+      "size": 5678,
+      ...
+    }
+  }
+]
+```
+
+Each object in the array contains:
+- `name`: The name of the file
+- `data`: The content of the file
+- `attrs`: Object attributes from GCP Storage including metadata like creation time, size, content type, etc.
 
 ### Delete object
 
@@ -257,7 +328,7 @@ An HTTP 204 (No Content) and empty body will be retuned if successful.
 
 ### List objects
 
-To perform a list object operation, invoke the S3  binding with a `POST` method and the following JSON body:
+To perform a list object operation, invoke the GCP bucket binding with a `POST` method and the following JSON body:
 
 ```json
 {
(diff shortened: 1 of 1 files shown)


daprdocs/content/en/reference/components-reference/supported-bindings/gcpbucket.md
```diff
@@ -211,6 +216,72 @@ The metadata parameters are:
...
+      ...
+    }
+  }
+]
+```
+
+Each object in the array contains:
+- `name`: The name of the file
+- `data`: The content of the file
+- `attrs`: Object attributes from GCP Storage including metadata like creation time, size, content type, etc.
```
alicejgibbons, line 284: Can we list all of these out or do they changes depending ong the bucket @nelson-parente

daprdocs/content/en/reference/components-reference/supported-bindings/gcpbucket.md
```diff
@@ -316,6 +387,58 @@ The list of objects will be returned as JSON array in the following form:
...
+To perform a copy object operation, invoke the GCP bucket binding with a `POST` method and the following JSON body:
+
+```json
+{
+  "operation": "copy",
+  "metadata": {
+    "destinationBucket": "destination-bucket-name",
+  }
+}
+```
```
alicejgibbons, line 402: Do we need specific responses on any of these? Assuming 204 no content for success or?

alicejgibbons: maybe just link to the output binding ref: https://docs.dapr.io/reference/api/bindings_api/#http-response-codes-2

nelson-parente:
added label back since it is still waiting on: https://github.com/dapr/components-contrib/pull/3810

alicejgibbons:
Code pr is merged here, just need to update the docs based on my feedback @nelson-parente

daprdocs/content/en/reference/components-reference/supported-bindings/gcpbucket.md
nelson-parente: we could but the list is pretty big 🙁 

```
// ObjectAttrs represents the metadata for a Google Cloud Storage (GCS) object.
type ObjectAttrs struct {
	// Bucket is the name of the bucket containing this GCS object.
	// This field is read-only.
	Bucket string

	// Name is the name of the object within the bucket.
	// This field is read-only.
	Name string

	// ContentType is the MIME type of the object's content.
	ContentType string

	// ContentLanguage is the content language of the object's content.
	ContentLanguage string

	// CacheControl is the Cache-Control header to be sent in the response
	// headers when serving the object data.
	CacheControl string

	// EventBasedHold specifies whether an object is under event-based hold. New
	// objects created in a bucket whose DefaultEventBasedHold is set will
	// default to that value.
	EventBasedHold bool

	// TemporaryHold specifies whether an object is under temporary hold. While
	// this flag is set to true, the object is protected against deletion and
	// overwrites.
	TemporaryHold bool

	// RetentionExpirationTime is a server-determined value that specifies the
	// earliest time that the object's retention period expires.
	// This is a read-only field.
	RetentionExpirationTime time.Time

	// ACL is the list of access control rules for the object.
	ACL []ACLRule

	// If not empty, applies a predefined set of access controls. It should be set
	// only when writing, copying or composing an object. When copying or composing,
	// it acts as the destinationPredefinedAcl parameter.
	// PredefinedACL is always empty for ObjectAttrs returned from the service.
	// See https://cloud.google.com/storage/docs/json_api/v1/objects/insert
	// for valid values.
	PredefinedACL string

	// Owner is the owner of the object. This field is read-only.
	//
	// If non-zero, it is in the form of "user-<userId>".
	Owner string

	// Size is the length of the object's content. This field is read-only.
	Size int64

	// ContentEncoding is the encoding of the object's content.
	ContentEncoding string

	// ContentDisposition is the optional Content-Disposition header of the object
	// sent in the response headers.
	ContentDisposition string

	// MD5 is the MD5 hash of the object's content. This field is read-only,
	// except when used from a Writer. If set on a Writer, the uploaded
	// data is rejected if its MD5 hash does not match this field.
	MD5 []byte

	// CRC32C is the CRC32 checksum of the object's content using the Castagnoli93
	// polynomial. This field is read-only, except when used from a Writer or
	// Composer. In those cases, if the SendCRC32C field in the Writer or Composer
	// is set to is true, the uploaded data is rejected if its CRC32C hash does
	// not match this field.
	//
	// Note: For a Writer, SendCRC32C must be set to true BEFORE the first call to
	// Writer.Write() in order to send the checksum.
	CRC32C uint32

	// MediaLink is an URL to the object's content. This field is read-only.
	MediaLink string

	// Metadata represents user-provided metadata, in key/value pairs.
	// It can be nil if no metadata is provided.
	//
	// For object downloads using Reader, metadata keys are sent as headers.
	// Therefore, avoid setting metadata keys using characters that are not valid
	// for headers. See https://www.rfc-editor.org/rfc/rfc7230#section-3.2.6.
	Metadata map[string]string

	// Generation is the generation number of the object's content.
	// This field is read-only.
	Generation int64

	// Metageneration is the version of the metadata for this
	// object at this generation. This field is used for preconditions
	// and for detecting changes in metadata. A metageneration number
	// is only meaningful in the context of a particular generation
	// of a particular object. This field is read-only.
	Metageneration int64

	// StorageClass is the storage class of the object. This defines
	// how objects are stored and determines the SLA and the cost of storage.
	// Typical values are "STANDARD", "NEARLINE", "COLDLINE" and "ARCHIVE".
	// Defaults to "STANDARD".
	// See https://cloud.google.com/storage/docs/storage-classes for all
	// valid values.
	StorageClass string

	// Created is the time the object was created. This field is read-only.
	Created time.Time

	// Deleted is the time the object was deleted.
	// If not deleted, it is the zero value. This field is read-only.
	Deleted time.Time

	// Updated is the creation or modification time of the object.
	// For buckets with versioning enabled, changing an object's
	// metadata does not change this property. This field is read-only.
	Updated time.Time

	// CustomerKeySHA256 is the base64-encoded SHA-256 hash of the
	// customer-supplied encryption key for the object. It is empty if there is
	// no customer-supplied encryption key.
	// See // https://cloud.google.com/storage/docs/encryption for more about
	// encryption in Google Cloud Storage.
	CustomerKeySHA256 string

	// Cloud KMS key name, in the form
	// projects/P/locations/L/keyRings/R/cryptoKeys/K, used to encrypt this object,
	// if the object is encrypted by such a key.
	//
	// Providing both a KMSKeyName and a customer-supplied encryption key (via
	// ObjectHandle.Key) will result in an error when writing an object.
	KMSKeyName string

	// Prefix is set only for ObjectAttrs which represent synthetic "directory
	// entries" when iterating over buckets using Query.Delimiter. See
	// ObjectIterator.Next. When set, no other fields in ObjectAttrs will be
	// populated.
	Prefix string

	// Etag is the HTTP/1.1 Entity tag for the object.
	// This field is read-only.
	Etag string

	// A user-specified timestamp which can be applied to an object. This is
	// typically set in order to use the CustomTimeBefore and DaysSinceCustomTime
	// LifecycleConditions to manage object lifecycles.
	//
	// CustomTime cannot be removed once set on an object. It can be updated to a
	// later value but not to an earlier one. For more information see
	// https://cloud.google.com/storage/docs/metadata#custom-time .
	CustomTime time.Time

	// ComponentCount is the number of objects contained within a composite object.
	// For non-composite objects, the value will be zero.
	// This field is read-only.
	ComponentCount int64

	// Retention contains the retention configuration for this object.
	// ObjectRetention cannot be configured or reported through the gRPC API.
	Retention *ObjectRetention

	// SoftDeleteTime is the time when the object became soft-deleted.
	// Soft-deleted objects are only accessible on an object handle returned by
	// ObjectHandle.SoftDeleted; if ObjectHandle.SoftDeleted has not been set,
	// ObjectHandle.Attrs will return ErrObjectNotExist if the object is soft-deleted.
	// This field is read-only.
	SoftDeleteTime time.Time

	// HardDeleteTime is the time when the object will be permanently deleted.
	// Only set when an object becomes soft-deleted with a soft delete policy.
	// Soft-deleted objects are only accessible on an object handle returned by
	// ObjectHandle.SoftDeleted; if ObjectHandle.SoftDeleted has not been set,
	// ObjectHandle.Attrs will return ErrObjectNotExist if the object is soft-deleted.
	// This field is read-only.
	HardDeleteTime time.Time
}

daprdocs/content/en/reference/components-reference/supported-bindings/gcpbucket.md
nelson-parente: added response sections!

msfussell:
@nelson-parente - Presuming that you are still doing updates to this PR, is that correct?

nelson-parente:
@msfussell I believe its good to go from my side!

alicejgibbons (approved):

msfussell (approved):

PR #4854, opened 2025-09-02
[1.16] Go Workflow durabletask

JoshVanL:
The Go SDK workflow wrapper over durabletask-go is to be deprecated, in favour of using the durabletask-go client directly. The Go SDK wrapper is unnecessary indirection, and prevents some features from being used by the user.


msfussell (approved):
LGTM!

Issue #4728, opened 2025-07-16
Multi-application Workflow


cicoyle:
cross app workflows

- [ ] call activity
- [ ] suborchestration

msfussell:
Closing. Resolved

PR #4855, opened 2025-09-02
[1.16] Workflow: Multi App

JoshVanL:
Adds documentation for multi application workflows.

Adds examples for Go and Java (but only activities).

Adds a section on App ID identity which is referenced.

Part of https://github.com/dapr/docs/issues/4728


msfussell (approved):
LGTM

PR #4865, opened 2025-09-08
[1.16] Adds workflow limitations

JoshVanL:
Updates state store table to include supported stores.

Adds a warning emoji to state stores which have workflow limitations.

Adds workflow limitation sections to the cosmosDB & dynamodb setup docs.


daprdocs/content/en/reference/components-reference/supported-state-stores/setup-azure-cosmosdb.md
```diff
@@ -225,6 +225,14 @@ This particular optimization only makes sense if you are saving large objects to
 
 {{% /alert %}}
 
+## Workflow Limitations
+
+The more complex a workflow is (number of activities, child workflows, etc.), the more state operations it will perform per state store transaction.
+All input & output values will also be saved to the workflow history, and will be part of an operation of these transactions.
+CosmosDB has a [maximum document size of 2MB and maximum transaction size of 100 operations.](https://learn.microsoft.com/en-us/azure/cosmos-db/concepts-limits#per-request-limits).
+This means that the workflow history must not exceed this size, meaning that CosmosDB is not suitable for workflows with large input/output values or larger complex workflows.
+
```
msfussell, line 234: @JoshVanL - can we provide this info ?
1) A clear description of the error thrown when this limit is reached and how do I handle this either in my code or by looking at the logs
2) Some way that I can provide a simple calculation. For example can we say that typically a workflow will need to stay below X number of activities etc.

daprdocs/content/en/reference/components-reference/supported-state-stores/setup-azure-cosmosdb.md
```diff
@@ -225,6 +225,14 @@ This particular optimization only makes sense if you are saving large objects to
 
 {{% /alert %}}
 
+## Workflow Limitations
+
+The more complex a workflow is (number of activities, child workflows, etc.), the more state operations it will perform per state store transaction.
```
msfussell:
```suggestion
The more complex a workflow is with number of activities, child workflows, etc, the more DB state operations it performs per state store transaction.
```

daprdocs/content/en/reference/components-reference/supported-state-stores/setup-azure-cosmosdb.md
```diff
@@ -225,6 +225,14 @@ This particular optimization only makes sense if you are saving large objects to
...
+All input & output values will also be saved to the workflow history, and will be part of an operation of these transactions.
```
msfussell:
```suggestion
All input & output values are saved to the workflow history, and are part of an operation of these transactions.
```

msfussell:
@JoshVanL - Also looks like there is a failing link in the CosmosDB file. Should not have localization. 
Also suggest that we update this https://v1-16.docs.dapr.io/developing-applications/building-blocks/workflow/workflow-overview/#limitations to point to the state stores with limitations as a list.

JoshVanL:
@msfussell PTAL

olitomlinson:
Just my 2c, I don't believe we should introduce a **limited-support** status for Workflows. It should either be a simple **yes** or **no**. 

I can easily foresee some users adopting Workflows (on limited components), which may work initially for small use-cases, but then over the course of time usage increases and then they start bumping up against these limitations in production, causing unexpected problems, which is never good for building the credibility of Dapr Workflows. 

What would our response be to this? _"We told you so. RTM"_ Nope. Thats not a good look IMO. We don't have any tooling to allow users to port their operational data from one State Store to the other, so we have no easy off-ramp from a limited-support State Store onto one that is fully supported. Not good.

Some of these limitations are immovable, like CosmosDb for example. It feels like we're setting up users for future failure by leading them down a path that we know we can't support them on, rather than allowing users to fall into the pit of success.

_Thought experiment_ : Name one other **stable** Building Block where we give mixed signals about its ability to adhere to the Abstraction of what the Building Block provides, on a per component basis? I don't believe we do? So why are we starting now? Happy to be wrong of course.

JoshVanL:
@olitomlinson I disagree- like all software, there are limitations on what it can do based on the technology or hardware. This is is true for all systems, so I think it is entirely appropriate that these limitations are exposed to the user. A user may already be using one of these technologies, and their use case entirely fits into these limitations. By not offering this as a store, we unnecessarily scope down Dapr's feature set.
The same argument can be made for using CosmosDB directly, entirely separate from Dapr.

The state store building block has all kinds of limitations on API, and again, the raw limitation of the db technology again.

olitomlinson:
> like all software, there are limitations on what it can do based on the technology or hardware

Sure, there are always limits, but this is a sneaky one! 

My main concern is that there is no off-ramp here. Workflows has taken enough hammer over the last few months for not having good operational tooling, and now we're handing over another footgun.

> By not offering this as a store, we unnecessarily scope down Dapr's feature set.

I don't see how we're scoping down the feature set? The features stay the same. We're limiting the audience for sure though.

> The state store building block has all kinds of limitations on API, and again, the raw limitation of the db technology again.

Yeah, having CosmosDb as a **transactional** compliant State Store component, but has a low upper bound of 100, is a mess, and I'm sure knowing what we know now we would make stricter requirements on what constitutes **transaction** support (see Redis)

We're always going to hit unknown unknowns, its unavoidable. But here we are making a decision on something we absolutely do know about. 

Sure adopters use-cases might be fine today, but what about then they want to do _more_ with Workflows? We can't support two actor state stores being online at one time, so customers will have to do a full offline migration, for which we offer no operational tools. It's not a good outlook.

JoshVanL:
@olitomlinson I have remove the warning from the table to instead mark dynamoDB and cosmosDB as not supported, but have left the verbiage in the components docs pages respectively.
Do you think this is a good middle-ground?

daprdocs/content/en/developing-applications/building-blocks/workflow/workflow-architecture.md
```diff
@@ -175,6 +175,20 @@ Similarly, if a state store imposes restrictions on the size of a batch transact
 Workflow state can be purged from a state store, including all its history.
 Each Dapr SDK exposes APIs for purging all metadata related to specific workflow instances.
 
+#### State store record count
+
+The number of records which will be saved as history in the state store per workflow run is determined by its complexity or "shape", i.e. the number of activities, timers, sub-workflows etc.
```
msfussell:
```suggestion
The number of records which are saved as history in the state store per workflow run is determined by its complexity or "shape". In other words, the number of activities, timers, sub-workflows etc.
```

daprdocs/content/en/reference/components-reference/supported-state-stores/setup-azure-cosmosdb.md
```diff
@@ -225,6 +225,15 @@ This particular optimization only makes sense if you are saving large objects to
 
 {{% /alert %}}
 
+## Workflow Limitations
+
+The more complex a workflow is with number of activities, child workflows, etc, the more DB state operations it performs per state store transaction.
+All input & output values are saved to the workflow history, and are part of an operation of these transactions.
+CosmosDB has a [maximum document size of 2MB and maximum transaction size of 100 operations.](https://learn.microsoft.com/azure/cosmos-db/concepts-limits#per-request-limits).
+Attempting to write to CosmosDB beyond these limits will result in an error code of `413`.
```
msfussell:
```suggestion
Attempting to write to CosmosDB beyond these limits results in an error code of `413`.
```

daprdocs/content/en/reference/components-reference/supported-state-stores/setup-azure-cosmosdb.md
```diff
@@ -225,6 +225,15 @@ This particular optimization only makes sense if you are saving large objects to
 
 {{% /alert %}}
 
+## Workflow Limitations
+
+The more complex a workflow is with number of activities, child workflows, etc, the more DB state operations it performs per state store transaction.
```
msfussell, line 237: Suggesting adding here that CosmosDB is not recommended for production when using workflow. WDYT?

daprdocs/content/en/reference/components-reference/supported-state-stores/setup-dynamodb.md
```diff
@@ -158,6 +158,13 @@ $ aws dynamodb get-item \
 }
 ```
 
+## Workflow Limitations
+
+The more complex a workflow is (number of activities, child workflows, etc.), the more state operations it will perform per state store transaction.
```
msfussell: Suggesting adding here that DynamoDB is not recommended for production when using workflow. WDYT?

daprdocs/content/en/reference/components-reference/supported-state-stores/setup-dynamodb.md
```diff
@@ -158,6 +158,13 @@ $ aws dynamodb get-item \
...
+The maximum number of operations that can be performed by DynamoDB in a [single transaction is 100](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/transaction-apis.html).
+This means that DynamoDB can only hadle workflows with a limited complexity, meaning it is not suitable for all workflow scenarios.
```
msfussell:
```suggestion
This means that DynamoDB can only handle workflows with a limited complexity, meaning it is not suitable for all workflow scenarios.
```

daprdocs/content/en/reference/components-reference/supported-state-stores/setup-dynamodb.md
msfussell:
```suggestion
The more complex a workflow is (number of activities, child workflows, etc.), the more state operations it performs per state store transaction.
```

daprdocs/layouts/_partials/components/state-stores.html
```diff
@@ -67,6 +67,10 @@ <h3>{{ $group }}</h3>
         <td align="center">
             {{ if .features.workflow }}
             <span role="img" aria-label="Workflow: Supported">✅</span>
+            {{ else if .features.workflow_warning }}
```
msfussell: Keep this or make this "false" not supported?

JoshVanL:
@msfussell updated, PTAL

daprdocs/content/en/reference/components-reference/supported-state-stores/setup-azure-cosmosdb.md
```diff
@@ -225,6 +225,22 @@ This particular optimization only makes sense if you are saving large objects to
 
 {{% /alert %}}
 
+## Workflow Limitations
+
+{{% alert title="Note" color="primary" %}}
+
+As described below, CosmosDB has limitations that likely make it unsuitable for production environments.
+There is currently no path for migrating Workflow data from CosmosDB to another state store, meaning exceeding these limits in production will result in failed workflows this no workaround.
```
msfussell:
```suggestion
There is currently no path for migrating Workflow data from CosmosDB to another state store, meaning exceeding these limits in production will result in failed workflows with no workaround.
```

daprdocs/content/en/reference/components-reference/supported-state-stores/setup-dynamodb.md
```diff
@@ -158,6 +158,20 @@ $ aws dynamodb get-item \
 }
 ```
 
+## Workflow Limitations
+
+
+As described below, DynamoDB has limitations that likely make it unsuitable for production environments.
+There is currently no path for migrating Workflow data from DynamoDB to another state store, meaning exceeding these limits in production will result in failed workflows this no workaround.
```
msfussell:
```suggestion
There is currently no path for migrating Workflow data from DynamoDB to another state store, meaning exceeding these limits in production will result in failed workflows with no workaround.
```

msfussell (commented):
Merged two grammar issues. @olitomlinson - Any comment?

msfussell (approved):
LGTM

PR #4904, opened 2025-09-25
Removing wait_until_ready, it has been renamed for consistency

acroca:
Ref: https://github.com/dapr/python-sdk/pull/843

`wait_until_ready` has been removed from the python SDK, and it's the only reference I could find to that method ([github search](https://github.com/search?q=org%3Adapr%20wait_until_ready&type=code)).

It has been renamed to `wait_for_sidecar`, which has a different casing pattern than the documented, but I think it's fine.


msfussell (approved):
LGTM

marcduiker (approved):
LGTM

PR #4906, opened 2025-10-01
Change max-body-size to use unit in documentation

jancespivo:
According to the code change and particularly test here https://github.com/dapr/dapr/blob/c71b329a2249e70cc0f38ae88af7d50abcd28a4f/pkg/injector/patcher/sidecar_container_test.go#L1331 The flag gets `string` not `int` as the original `http-max-request-size` and requires unit specification, otherwise bytes unit is used as a default.


msfussell (approved):

marcduiker:
@holopin-bot @jancespivo Here's a digital badge as a small token of appreciation.

PR #4909, opened 2025-10-03
Adds documentation for `$ dapr scheduler`

JoshVanL:
See https://github.com/dapr/cli/pull/1559


marcduiker (approved):
LGTM

daprdocs/content/en/concepts/dapr-services/scheduler.md
```diff
@@ -115,25 +115,85 @@ services:
       - ./dapr_scheduler/2:/var/run/dapr/scheduler
 ```
 
-## Back Up and Restore Scheduler Data
+## Managing jobs with the Dapr CLI
```
cicoyle, line 118: My nit would be I think we should have a section on job naming convention here since it differs slightly across 'job types'. 

For example, we have actor reminders prefixed with `actor/type/id/name`, jobs are prefixed with `app/...` and wf related 'jobs' due to the reminders being used under the hood are `workflow/...` prefixed vs `activity/...` prefixed. I think we should just spell this out very explicitly to limit confusion to users.

daprdocs/content/en/concepts/dapr-services/scheduler.md
```diff
@@ -115,25 +115,85 @@ services:
...
+Delete one or more specific jobs:
+
+```bash
+dapr scheduler delete app/my-app/job1 actor/MyActor/123/reminder1
+```
+
+Bulk delete jobs with filters:
+
+```bash
+dapr scheduler delete-all all
```
cicoyle, line 187: Was using a wildcard (`*`) considered here? so it could be:

```
dapr scheduler delete *
dapr scheduler delete app/*
dapr scheduler delete actor/*
etc
```
A single quote (`''`) around the name might be required with the `*`.

This would prevent us from have a `delete-all` flag/subcommand, as users could delete bulk by passing the wildcard. Or even we could have `dapr scheduler delete --pattern 'app/'` which would also reduce the need of the `delete-all` subcommand.

daprdocs/content/en/developing-applications/building-blocks/actors/actors-timers-reminders.md
```diff
@@ -187,6 +187,66 @@ To use protobuf serialization for actor reminders on self-hosted, use the follow
 --max-api-level=20
 ```
 
+## Managing reminders with the CLI
```
cicoyle, line 190: My pref would be if we could move this to the `scheduler.md` and make a note here linking to the scheduler page so all management of all job types with the cli can be seen from the one scheduler page since scheduler stores them all. Here we should still have this section but simply link to the `scheduler.md` so all the info is in one place.

Or if ^ is not preferred, can we at least link to this documentation from the `scheduler.md` saying something to the tune of `please see here for how to manage reminders with the CLI` bc I could totally see people going to the scheduler docs for this information too.

daprdocs/content/en/developing-applications/building-blocks/actors/actors-timers-reminders.md
```diff
@@ -187,6 +187,66 @@ To use protobuf serialization for actor reminders on self-hosted, use the follow
...
+dapr scheduler delete-all actor/MyActorType
+```
+
+Delete all reminders for a specific actor instance:
+
+```bash
+dapr scheduler delete-all actor/MyActorType/actorid1
+```
+
+#### Backup and restore reminders
```
cicoyle, line 230: if we have all docs from the `scheduler.md` if would limit this being noted twice bc this is the same as: `Export all jobs to a file:`

cicoyle (changes requested):
Generally this is great and will be super beneficial to users. Just some ideas around consolidating the docs into one place and linking it across the other reference sections to limit sprawl & duplication of the same information. 

Also had a question around the design of the `delete-all` - might be worth commenting on the PR in the cli, but I reviewed this first for a better user experience perspective.

daprdocs/content/en/concepts/dapr-services/scheduler.md
JoshVanL: We went with the `delete-all` command for v1.16

daprdocs/content/en/developing-applications/building-blocks/actors/actors-timers-reminders.md
JoshVanL: :+1:

daprdocs/content/en/developing-applications/building-blocks/actors/actors-timers-reminders.md
JoshVanL: I hear you, however I think I prefer to still have this here too

cicoyle (approved):
fine by me. thx!

nelson-parente (approved):
LGTM @marcduiker can we merge it?

PR #4994, opened 2026-01-06
[1.17] Remove reference to SchedulerReminders feature gate

JoshVanL:
As of 1.17, Dapr no longer supports the `SchedulerReminders` feature gate meaning all actor reminders are managed via Scheduler.

PR removes all reference to it, including documentation on reminder migration & old configuration for the legacy "state" reminders.


msfussell (approved):
LGTM. Great change we should highlight in the release notes.

marcduiker (approved):
LGTM

PR #5074, opened 2026-03-17
workflow: retention policy is not retroactive

JoshVanL:
Clarify that adding or changing a retention policy does not retroactively purge workflows already in a terminal state. Include CLI examples for retroactive cleanup using `dapr workflow purge --all-older-than` and `dapr workflow list --filter-status` combined with purge.


marcduiker (approved):
LGTM

PR #5075, opened 2026-03-17
Workflow: `wf purge --all-older-than --all-filter-status`

JoshVanL:
Update docs based on new CLI flag https://github.com/dapr/cli/pull/1609


marcduiker (approved):
LGTM

PR #5077, opened 2026-03-19
[1.18] Azure blob storage bulk

JoshVanL:
- bulkGet: parallel blob retrieval with streaming-to-file and inline (base64) modes, prefix-based discovery, and configurable concurrency
- bulkCreate: parallel blob upload from local files or inline data, with per-item contentType support
- bulkDelete: parallel blob deletion by name or prefix, using the Azure Blob Batch API (256-key chunks) with individual fallback

Implemented in https://github.com/dapr/components-contrib/pull/4295


marcduiker (approved):
LGTM

nelson-parente (commented):
Pre-review for @marcduiker handoff: LGTM — large but well-structured. Already approved.
Documents three new Azure Blob Storage bulk operations: `bulkGet` (streaming to files or inline base64, prefix-based discovery, configurable concurrency), `bulkCreate` (parallel upload from local files or inline, per-item contentType), `bulkDelete` (by name or prefix, Batch API with individual fallback). Security note about path traversal prevention on streaming mode is present. All operations have Windows/Linux tabpane curl examples. Safe to merge after CI.

PR #5078, opened 2026-03-19
[1.18] pubsub/kafka: per message partitioning

JoshVanL:
Document the new partitionNumber per-call metadata field, including usage example, precedence rules over partitionKey, and valid values.

Implemented in https://github.com/dapr/components-contrib/pull/4298


marcduiker (approved):
LGTM

nelson-parente (commented):
Pre-review for @marcduiker handoff: LGTM. Already approved. Documents per-message partition number for Kafka pub/sub: `partitionNumber` metadata takes precedence over `partitionKey`/`__key`, valid range is 0 to partitionCount-1, and the message headers section is updated to exclude `partitionNumber`. The curl example is correct. Safe to merge after CI.

PR #5079, opened 2026-03-19
[1.18] bindings/azure/blobstorage: SAS URL support

JoshVanL:
Implemented in https://github.com/dapr/components-contrib/pull/4299


marcduiker (approved):
LGTM

nelson-parente (commented):
Pre-review for @marcduiker handoff: LGTM. Already approved. Documents Azure Blob Storage SAS URL support: new `presign` operation with full request/response examples, `signTTL` field added to the create operation for generating a URL at upload time, and the important note that account key or connection string is required (Entra ID not supported). Windows/Linux tabpane usage is consistent. Safe to merge after CI.

PR #5144, opened 2026-04-22
[1.18] update sidecar & control plane probe defaults

JoshVanL:
See https://github.com/dapr/dapr/pull/9818


nelson-parente (commented):
Pre-review for @marcduiker handoff: LGTM. Sidecar probe defaults updated for v1.18: liveness initialDelay 3s → 180s, period 3s → 10s; readiness initialDelay 3s → 1s, period 3s → 1s, timeout 3s → 1s, failureThreshold 3 → 5. Both the YAML examples and the annotations reference table are consistently updated. The `/v1.0/healthz` path now correctly includes the leading slash. Safe to merge after CI.

marcduiker (approved):
LGTM

PR #5168, opened 2026-05-19
[1.18] workflow: document payload-size graceful stall and ratio metrics

JoshVanL:
New how-to page covering the v1.18 changes from dapr/dapr#9847 and dapr/dapr#9929:

- Why the stream tear-down used to happen and how the 95%-of-max-body-size pre-check now stalls only the offending workflow.
- Remediation options: raise --max-body-size and restart, force-purge, or restructure to pass references instead of inline data.
- The two new histogram metrics runtime/workflow/payload/size_ratio and runtime/workflow/activity/payload/size_ratio, their tags, their buckets, and sample Prometheus alerts that work independent of the configured --max-body-size.

Add a short pointer from workflow-features-concepts.md Limitations to the new page so it is discoverable from the main concepts doc.


nelson-parente (commented):
Pre-review for @marcduiker handoff: LGTM. New page `workflow-payload-size.md` with companion pointer in workflow-features-concepts.md. Content is technically solid: the 95% budget, graceful stall semantics, the two ratio metrics with tags and bucket values, and the Prometheus alert templates. The PromQL queries are correct and correctly note they're independent of the actual `--max-body-size` value. Safe to merge after CI.

msfussell (commented):
@JoshVanL - Can you look at the CoPilot suggestion. 
Also I presume you also need to update this https://github.com/dapr/dapr/blob/master/docs/development/dapr-metrics.md with the new metrics.

JoshVanL:
@msfussell @nelson-parente I have updated-  please take another look. 

I have also opened this: https://github.com/dapr/dapr/pull/10005

msfussell (approved):
LGTM @nelson-parente

PR #5169, opened 2026-05-19
[1.18] actors: document drain timeout clamping to dissemination budget

JoshVanL:
Cover dapr/dapr#9897:

- drainOngoingCallTimeout is clamped at runtime when it meets or exceeds the placement dissemination timeout, to avoid daprd resetting the placement stream during drain.
- Effective value falls back to 80% of the dissemination timeout, with a floor at the runtime DefaultOngoingCallTimeout, and a warning is logged.
- Per-actor-type drainOngoingCallTimeout values set under entitiesConfig are honored and subject to the same clamp.

Update both the runtime config table and add a Drain timeout clamping section explaining when the clamp kicks in and how operators should respond.


nelson-parente (commented):
Pre-review for @marcduiker handoff: LGTM. Documents the drain timeout clamping behaviour (v1.18): the 80% dissemination-budget floor is explained clearly, the per-entity-type scope is noted, and the table entry for `drainOngoingCallTimeout` is updated with a cross-link to the new section. The warning to operators about what to do when they see the log message is a nice touch. Safe to merge after CI.

marcduiker (approved):
LGTM

PR #5170, opened 2026-05-19
[1.18] api: document the Shutdown API and Dapr-Force-Shutdown option

JoshVanL:
New reference page for the Shutdown HTTP and gRPC API. Covers:

- The pre-1.18 bug where the endpoint was a no-op because the handler waited on a request context Done channel that does not fire after the response is sent. dapr/dapr#9949 fixes this so the endpoint now reliably triggers graceful shutdown.
- The new Dapr-Force-Shutdown HTTP header and dapr-force-shutdown gRPC metadata key, which skip graceful shutdown and exit via os.Exit(1) after a short delay (intended for wedged sidecars and end-to-end test teardown).
- When to use graceful vs force, including the exit-code-1 caveat.

Also link the Shutdown row in api-allowlist.md to the new reference page.


nelson-parente (commented):
Pre-review for @marcduiker handoff: LGTM. New reference page for the Shutdown API is accurate and complete: graceful vs. force modes, HTTP/gRPC examples, the `Dapr-Force-Shutdown` header and gRPC metadata key, exit-code-1 caveat, and the api-allowlist cross-link. The prose correctly notes the pre-1.18 bug (no-op due to request context Done channel) and that dapr/dapr#9949 fixes it. Safe to merge after CI.

daprdocs/content/en/reference/api/shutdown_api.md
```diff
@@ -0,0 +1,76 @@
+---
+type: docs
+title: "Shutdown API reference"
+linkTitle: "Shutdown API"
+description: "Detailed documentation on the Shutdown API"
+weight: 1100
```
marcduiker:
```suggestion
weight: 1150
```

marcduiker (approved):

PR #4935, opened 2025-11-03
docs: Clarify max-body-size applies to requests and responses (v1.16.3)

jukhancisco:
## Summary

This PR clarifies the documentation for the `dapr.io/max-body-size` annotation to explicitly state that it applies to both request and response bodies, not just incoming requests.

**Target**: v1.16 documentation (for Dapr v1.16.3 release)

## Changes

Updated the `max-body-size` annotation description in the arguments and annotations reference to clarify that it applies to:
- Incoming HTTP/gRPC requests to the Dapr sidecar
- Actor operation responses  
- **Binding component responses** (added in Dapr v1.16.3)

## Previous Description

> Inreasing the request max body size to handle large file uploads using http and grpc protocols. Set the value using size units (e.g., `16Mi` for 16MB). The default is `4Mi`

## New Description

> Maximum size for HTTP/gRPC request and response bodies. Applies to incoming requests to the Dapr sidecar, actor operations, and binding component responses. Useful for handling large file uploads/downloads. Set the value using size units (e.g., `16Mi` for 16MB). The default is `4Mi`. Note: Binding component response size enforcement added in Dapr v1.16.3.

## Why This Change?

Users were confused about the scope of the `max-body-size` annotation. The fix in PR https://github.com/dapr/dapr/pull/9151 (targeting v1.16.3) ensures binding components also respect this annotation, fixing issue https://github.com/dapr/dapr/issues/9177.

This documentation update helps users understand:
1. The annotation controls both requests AND responses
2. It applies to binding component responses (as of v1.16.3)
3. It's useful for both uploads and downloads

## Related Issues/PRs

- Fixes: https://github.com/dapr/dapr/issues/9177
- Related: https://github.com/dapr/dapr/pull/9151 (v1.16.3 bug fix)

## Checklist

- [x] Updated annotation description for v1.16 docs
- [x] Referenced the version where binding support was added (v1.16.3)
- [x] Clarified scope (requests + responses)
- [x] Linked related issues/PRs
- [x] Targeted v1.16 documentation branch

PR #5185, opened 2026-05-25
docs: Clarify max-body-size applies to requests and responses

JoshVanL:
Updated the max-body-size annotation description to clarify that it applies to:
- Incoming HTTP/gRPC requests to the Dapr sidecar
- Actor operation responses
- Binding component responses (added in v1.16.2)

This clarification helps users understand that the annotation controls both request and response body sizes, not just incoming requests.

Forward port of https://github.com/dapr/docs/pull/4935/commits


msfussell (approved):
LGTM

PR #5195, opened 2026-06-01
docs: update our supported version docs to be correct

sicoyle:

- [X] Commits are signed with Developer Certificate of Origin (DCO - [learn more](https://docs.dapr.io/contributing/contributing-overview/#developer-certificate-of-origin-signing-your-work))
- [X] [Read the contribution guide](https://docs.dapr.io/contributing/docs-contrib/contributing-docs/)
- [X] Commands include options for Linux, MacOS, and Windows within tabpane
- [ ] Places where multiple code/command options are given have tabpane

## Description

I brought this table up in the maintainers sync and we agreed to remove it as it easily becomes outdated, and it not actually accurate.


sicoyle:
cc @marcduiker and @msfussell

msfussell:
We should not remove this table and instead make it accurate, if things are wrong (which I believe they are not). Either way, the point of this table is to have a source of truth on supported releases, otherwise that is not recorded anywhere.  We can simplify this list (we have done that before) and remove the older releases (as they are in the earlier doc version) but we should not remove this.

msfussell:
Closing. We need to keep the table. Update in another PR

sicoyle:
> We should not remove this table and instead make it accurate, if things are wrong (which I believe they are not). Either way, the point of this table is to have a source of truth on supported releases, otherwise that is not recorded anywhere. We can simplify this list (we have done that before) and remove the older releases (as they are in the earlier doc version) but we should not remove this.

@msfussell This was discussed in the maintainers call, and folks agreed to remove the table. A few reasons: the rest of this page already documents the n-2 release support policy, so the table duplicates that rather than serving as the source of truth. It's also drifted — for example, it still references the dashboard, which we recently removed and is but one more manual step we must do in the release process that does fall by the wayside. More importantly, no SDK that I'm aware of has tests verifying the claims in this table, so right now nothing proves it's accurate. We're making changes in the Python SDK to actually cover that gap and can roll the same approach out to the other SDKs, but until that exists, keeping a table we can't validate gives a false sense of a source of truth.

PR #5199, opened 2026-06-03
Externalize the sidebar nav tree to reduce page size

marcduiker:
The Docsy left-navigation tree was being serialized into every one of the ~1,742 generated pages, adding roughly 256 KB of identical markup to each one. Because the site home is type: docs, every page rendered the exact same tree rooted at .Site.Home — so the same content was duplicated across the entire build. This PR renders that tree once per language to a standalone sidebar-nav.html fragment (via a new SidebarNav Hugo output format) and injects it into each page client-side.

 A project override of _partials/sidebar.html replaces the inlined tree with an empty #td-section-nav container carrying data-nav-src (the language's fragment URL, cache-busted by site version) and data-nav-mid (the page's active-item id), plus a small <noscript> block of top-level section links for no-JS users and crawlers. A new static script, /js/sidebar-nav.js, fetches the fragment, injects it, and applies active-item highlighting client-side — mirroring the approach Docsy itself ships for large sites. The fragment template copies Docsy's tree recursion under a uniquely-named dapr-sidebar-nav-section define so it never collides with the theme's internal version.

Validated with a clean build in the devcontainer (Hugo 0.147.9-extended): both fragments emit correctly (/sidebar-nav.html and /zh-hans/sidebar-nav.html), pages now carry the data-nav-src reference instead of the full tree, and the mobile toggle and right-hand TOC are unaffected. **Total build output dropped from 514 MB to ~248 MB — roughly a 50% reduction.**


msfussell (approved):
LGTM

PR #5213, opened 2026-06-15
Workflows: expand on concurrency limits

JoshVanL:
Expand on the workflow concurrency limits and when to use either flavour.
Switch the sections to favour global concurrency limits.


marcduiker (approved):
LGTM

PR #5220, opened 2026-06-17
docs: clarify duration format for Zeebe command binding

nelson-parente:
## Summary

- Adds a prominent alert note to the Output binding section stating that all duration fields require Go duration strings (e.g. `"30s"`, `"5m"`, `"1h30m"`) — ISO-8601 durations like `"PT1M"` are not supported.
- Updates the inline description for every affected duration field to explicitly state the required format: `publish-message.timeToLive`, `create-instance.requestTimeout`, `activate-jobs.timeout`, `activate-jobs.requestTimeout`, and `fail-job.retryBackOff`.

## Motivation

Users sending ISO-8601 durations or other ambiguous strings to Zeebe command binding fields were receiving a confusing `Invalid duration` error with no indication of the expected format. The code PR (dapr/components-contrib#4412) improves the error message; this PR ensures the docs are unambiguous so users can avoid the error entirely.

Code: dapr/components-contrib#4412
Relates to: dapr/components-contrib#4093

## Target

Dapr 1.19

## Test plan

- [ ] Verify alert block renders correctly in the Dapr docs site
- [ ] Confirm all five affected fields now document the Go duration string format


marcduiker (approved):
LGTM

PR #5228, opened 2026-06-29
Fix RabbitMQ pubsub exchangeKind allowed values (add direct, headers)

MyMirelHub:
## Description

The RabbitMQ pub/sub reference page lists only `fanout` and `topic` as allowed values for `exchangeKind`, but the component actually accepts **four** exchange kinds. From [`exchangeKindValid`](https://github.com/dapr/components-contrib/blob/master/pubsub/rabbitmq/metadata.go) in components-contrib:

```go
func exchangeKindValid(kind string) bool {
	return kind == amqp.ExchangeFanout || kind == amqp.ExchangeTopic || kind == amqp.ExchangeDirect || kind == amqp.ExchangeHeaders
}
```

This updates the `exchangeKind` row to also list `direct` and `headers`.

Docs for https://github.com/dapr/components-contrib/issues/4427


alicejgibbons (approved):
lgtm!

marcduiker (approved):
LGTM

PR #5232, opened 2026-07-02
perf: add version-agnostic performance results page

JoshVanL:
Render the Dapr performance results on the docs site from the charts and manifest published to the dapr/dapr perf-charts branch, for whichever Dapr version the docs set represents.

- Add the dapr-perf-results shortcode: read the minor version from params.version, fetch <minor>/manifest.json from the perf-charts branch (via jsDelivr) at build time, and render the throughput-per-resource tables and charts. Fails soft to a "not yet published" notice when a version has no results.
- Group the output so the page stays navigable: each API is a collapsible section (collapsed by default) with a scenario count, each scenario is collapsible with a chart count, and a "Jump to" nav links to each API and expands it on click.
- Add the perf-results page that calls the shortcode. Nothing is hard-coded per version; publishing a new run to perf-charts is all that is needed for the page to update.

### DIFF
--- a/daprdocs/content/en/operations/performance-and-scalability/perf-results.md
+++ b/daprdocs/content/en/operations/performance-and-scalability/perf-results.md
@@ -6,9 +6,12 @@ weight: 10000
 description: "Performance benchmarks and charts for Dapr APIs"
 ---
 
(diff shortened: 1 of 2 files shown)


marcduiker (approved):
Let's see 👀

PR #5244, opened 2026-07-13
Workflow docs: instance ID reuse requires the whole workflow tree be terminal

JoshVanL:
The workflow identity docs added for the 1.18 reuse policy removal stated that creating a workflow whose instance ID already exists is rejected in any state, including completed. That never matched the runtime: recreating an instance in a terminal state has been allowed since the workflow engine was introduced, and remains so.

Document the contract as enforced from 1.18.2: an instance ID is reusable only once the existing workflow and every child workflow it created, checked recursively, has reached a terminal state (COMPLETED, FAILED, or TERMINATED); otherwise the create request is rejected with a 409 conflict naming the child workflow that is not yet terminal. Purging a workflow (or a retention policy) frees its instance ID. During a rolling upgrade, children hosted by older sidecars are verified only for their own status, so the recursive guarantee applies once all sidecars run a version with the check.

Also update the breaking-changes page: the 1.18.0 removal entry now notes that SDK options still exposing the reuse policy (such as reuse_id_policy in the Python SDK) are deprecated and silently ignored by 1.18+ runtimes, and a new 1.18.2 entry records the terminal-tree requirement.

- [ ] [Read the contribution guide](https://docs.dapr.io/contributing/docs-contrib/contributing-docs/)
- [ ] Commands include options for Linux, MacOS, and Windows within tabpane
- [ ] Places where multiple code/command options are given have tabpane


marcduiker (approved):
LGTM

PR #5249, opened 2026-07-15
Document updating the trust bundle ConfigMap when rotating certs via kubectl

### LINKED ISSUE #5248: mTLS kubectl cert-update steps omit the dapr-trust-bundle ConfigMap
**Describe the issue**

The instructions for updating the root/issuer certificates via `kubectl` (under "Custom certificates (bring your own)") tell the user to edit only the `dapr-trust-bundle` **Secret**. The trust bundle also exists as a `dapr-trust-bundle` **ConfigMap** (holding the public root `ca.crt`). On startup, Sentry checks that the ConfigMap's `ca.crt` matches the Secret's; if they differ it treats the bundle as invalid, logs `Root and issuer certs not found: generating self signed CA`, and regenerates a fresh self-signed 1-year CA — overwriting the user's certificates.

The kubectl section never mentions the ConfigMap, so a user following it exactly ends up back on a Sentry-generated 1-year cert.

**URL of the docs**

https://docs.dapr.io/operations/security/mtls/#custom-certificates-bring-your-own

**Expected content**

The kubectl steps should note that the `dapr-trust-bundle` ConfigMap's `ca.crt` must also be updated to match the Secret (plain PEM, not base64), with a command to do so.

**Screenshots**

N/A

**Additional context**

This affects only the edge case where neither Helm nor the dapr CLI is used to rotate certificates — typically high-compliance environments (e.g. banks) that deploy Dapr via rendered/static manifests and rotate certs entirely by hand with `kubectl`. Helm (`--set-file dapr_sentry.tls.*`) and the CLI (`dapr mtls renew-certificate`) both update the Secret and the ConfigMap atomically, so they never mismatch — which is likely why the ConfigMap step was never documented.

Verified in `pkg/sentry/server/ca/kube.go` at v1.14.4 (`get()` returns not-found when `cm.Data["ca.crt"] != secret ca.crt`); the behavior is still present on `master`.

MyMirelHub:
## Description

The `kubectl` certificate-update steps under "Custom certificates (bring your own)" only update the `dapr-trust-bundle` **Secret**. The trust bundle also lives in a `dapr-trust-bundle` **ConfigMap** (public root cert), and Sentry validates the two as a pair: if the ConfigMap's `ca.crt` doesn't match the Secret's, Sentry discards the bundle and silently regenerates a self-signed 1-year CA, overwriting the user's certificates. This also breaks the section's zero-downtime promise whenever the root cert changes — even when signed with the same private key.

This PR restructures that passage into numbered steps (matching the sibling self-signed subsection):

1. Update the Secret (existing text, unchanged)
2. If the root cert changed, sync the same root into the ConfigMap (plain PEM, not base64) — with the command; issuer-only rotations under the existing root can skip this
3. A verification line after the Sentry restart with the exact log message that indicates the mismatch (`Root and issuer certs not found: generating self signed CA`), since the failure is otherwise silent

Only the manual-`kubectl` path is affected; Helm and `dapr mtls renew-certificate` update both objects atomically.

## Issue reference

Fixes #5248

## Checklist

- [x] Followed the [style and terminology guidance](https://docs.dapr.io/contributing/docs-contrib/contributing-docs/#styling-conventions)
- [x] Verified against the Sentry behavior in `pkg/sentry/server/ca/kube.go` (v1.14.4 and `master`)


marcduiker (approved):
LGTM

alicejgibbons (approved):
lgtm, ty!

PR #5273, opened 2026-08-03
Clarify retry status-code matching doesn't apply to bulk subscribe RETRY

alicejgibbons:
## Summary
- `matching.httpStatusCodes`/`matching.gRPCStatusCodes` only filter on the transport-level status code of the call to the application.
- A bulk subscribe handler's per-entry `RETRY` status is returned in the response body, not as an HTTP/gRPC status code, so it is not affected by `matching` and is always retried per the policy's `duration`/`maxRetries`/backoff.
- Added a note to the "Retry filter based on status codes" section of the retries overview page to call this out, with a link to the bulk subscribe docs.

## Test plan
- [x] Verified against the dapr/dapr runtime source (`pkg/resiliency/policy.go`, `pkg/runtime/subscription/bulkresiliency.go`, `pkg/runtime/subscription/postman/http/http.go`, `pkg/runtime/subscription/postman/grpc/grpc.go`): the `matching` filter only fires when the error is a `resiliency.CodeError`, and bulk subscribe's per-entry `RETRY` never produces one.
- [ ] Docs site preview build (Hugo) once CI runs

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/daprdocs/content/en/operations/resiliency/policies/retries/retries-overview.md
+++ b/daprdocs/content/en/operations/resiliency/policies/retries/retries-overview.md
@@ -132,12 +132,10 @@ spec:
           gRPCStatusCodes: "4,8-11,13,14" # retry gRPC status codes in these ranges and separate single codes.
 ```
 
-{{% alert title="Note" color="primary" %}}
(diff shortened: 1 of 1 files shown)


marcduiker (approved):
LGTM

PR #5277, opened 2026-08-05
[1.18] Hot reloading of the actor state store

JoshVanL:
Since Dapr 1.18.3, the actor state store is no longer excluded from hot reloading. The actors runtime now reconciles actor hosting with the configured store at runtime: adding a store enables actor hosting and the workflow APIs, removing it drains hosted actors until a store returns, and in-place updates (such as a rotated secret) swap the store instance without interrupting actor hosting.

Added in dapr/dapr#10309, backported to release-1.18 (v1.18.3) in dapr/dapr#10325.


marcduiker (approved):
LGTM

PR #5295, opened 2026-08-19
AMQP 1.0: rename the Solace-AMQP page and document the generic component

nelson-parente:
## Description

Renames the Solace-AMQP page to the generic AMQP 1.0 component page, and documents the addressing behaviour the component change introduces.

Companion to the components-contrib change that renames `pubsub.solace.amqp` to `pubsub.amqp`. The component speaks plain AMQP 1.0, but it was named, documented and defaulted for one vendor, so a user on Apache ActiveMQ Artemis had to declare a component called "solace" and switch off its addressing convention by hand.

| Type | Default prefixes | Status |
|---|---|---|
| `pubsub.amqp` | none — addresses by name | the documented component |
| `pubsub.solace.amqp` | `topic://` and `queue://` | deprecated, still registered, existing configs keep working |

## What changed

- **Renamed** `setup-solace-amqp.md` to `setup-amqp.md`. The old URL keeps working through `aliases`.
- **Retitled** to AMQP 1.0, and said so in the first line. Dapr also ships AMQP **0-9-1** as `pubsub.rabbitmq`, and the two are different protocols, so a bare "AMQP" misleads.
- **Rewrote the addressing section** around the new defaults. The `queue:` and `topic:` schemes select which prefix applies — they do not select a routing type by themselves, and with no prefixes configured `queue:orders` and `orders` are the same address.
- **Documented the broker configuration that makes `queue:` meaningful**, which was missing before. An Artemis acceptor takes `anycastPrefix` and `multicastPrefix`, and the component has to be given the same values.
- **Added a migration section** — the three lines a Solace user changes to move to `pubsub.amqp`.
- **Added an interop warning.** A prefix namespaces the address, so a Dapr publisher on `topic://orders` and a non-Dapr client on `orders` never meet. That is the actual defect behind dapr/dapr#10363.
- **Documented what happens when a subscriber returns an error.** The component answers with `modified` and `delivery-failed`, so the broker's redelivery-delay and max-delivery-attempts apply.
- **Added** the `backOff*` fields the component reads, and an Artemis container to test against.

## Corrections to the page that were already wrong

A review of the page against the component source found four claims that are not true, and they are fixed here:

- `consumerID` was documented as a competing-consumer group key and shown in the first example. **The component never reads it.**
- `caCert`, `clientCert` and `clientKey` were marked "Required for using TLS". None is required — `caCert` is only needed for a privately signed broker certificate, the client pair only for mutual TLS, and all three require the `amqps://` scheme.
- The addressing section told the reader to prefix a topic with `queue:` to use a queue, then explained eleven lines later that this does nothing under the defaults.
- The supported components table said "since 1.10", copied from the Solace row. `pubsub.amqp` does not exist in any released runtime.

## Note for reviewers

This supersedes the original version of this PR, which documented Artemis as a Solace component. Please review it as a fresh page rather than as a diff.

Do not merge before dapr/components-contrib#4602 lands, since the page documents `pubsub.amqp` as the type to use. The runtime registration is dapr/dapr#10502.

## Issue reference

Relates to dapr/dapr#10363.

### DIFF
--- a/daprdocs/content/en/reference/components-reference/supported-pubsub/setup-solace-amqp.md
+++ b/daprdocs/content/en/reference/components-reference/supported-pubsub/setup-solace-amqp.md
@@ -2,7 +2,7 @@
 type: docs
 title: "Solace-AMQP"
 linkTitle: "Solace-AMQP"
-description: "Detailed documentation on the Solace-AMQP pubsub component"
+description: "Detailed documentation on the Solace-AMQP pubsub component, which also supports other AMQP 1.0 brokers such as Apache ActiveMQ Artemis"
 aliases:
   - "/operations/components/setup-pubsub/supported-pubsub/setup-solace-amqp/"
 ---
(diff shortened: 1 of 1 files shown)


daprdocs/content/en/reference/components-reference/supported-pubsub/setup-amqp.md
```diff
@@ -0,0 +1,231 @@
...
+
+To set up AMQP pub/sub, create a component of type `pubsub.amqp`. See the [pub/sub broker component file]({{% ref setup-pubsub.md %}}) to learn how ConsumerID is automatically generated. Read the [How-to: Publish and Subscribe guide]({{% ref "howto-publish-subscribe.md#step-1-setup-the-pubsub-component" %}}) on how to create and apply a pub/sub configuration.
+
+```yaml
+apiVersion: dapr.io/v1alpha1
+kind: Component
+metadata:
+  name: amqp-pubsub
+spec:
+  type: pubsub.amqp
```
alicejgibbons, line 27: @nelson-parente can you confirm if this is backwards compatible change or not?

daprdocs/content/en/reference/components-reference/supported-pubsub/setup-amqp.md
```diff
@@ -0,0 +1,231 @@
...
+with `delivery-failed`. The broker redelivers it and counts the attempt, so the
+broker's own `redelivery-delay` and `max-delivery-attempts` settings apply and a
+message that always fails eventually reaches the dead-letter address.
+
+Redelivery pacing and the attempt limit are broker settings, not component
+settings. On ActiveMQ Artemis they live in `address-settings`.
+
+## Migrating from `pubsub.solace.amqp`
+
+{{% alert title="Deprecated" color="warning" %}}
```
alicejgibbons, line 178: I see it is backwards compatible. Do we need to call out this solace type is deprecated and will be released in two verisons?

PR #5317, opened 2026-09-16
Add graceful shutdown guidance for actor hosts on Kubernetes

### LINKED ISSUE #5316: Document graceful shutdown configuration for actor hosts on Kubernetes

The docs describe `dapr.io/block-shutdown-duration`, `dapr.io/graceful-shutdown-seconds`, `drainOngoingCallTimeout`, and `drainRebalancedActors` individually, but nothing explains how they interact with Kubernetes pod termination for actor hosts.

Kubernetes sends `SIGTERM` to the application container and the daprd sidecar at the same time. An idle application typically exits in well under a second, while the sidecar is still in its shutdown sequence. The sidecar then fails every actor deactivation with connection refused (`Failed to halt all actors...`), `OnDeactivateAsync` never runs, and until the sidecar exits and placement is re-disseminated, other sidecars keep routing calls to actors placed on the dying pod — a multi-second window of failed actor invocations on every ordinary rolling restart. The drain settings are silently ineffective in this configuration because the application is already gone when draining happens.

We hit this in production and verified the behavior experimentally on dapr 1.18.4: with defaults, every rolling restart produces refused deactivations and a multi-second actor black-hole; with the sidecar-outlives-app configuration described below, deactivations complete against a live app and the failure window disappears.

A production-guidelines section explaining the parallel-SIGTERM pitfall and the sidecar-outlives-app pattern: a `preStop` sleep on the application container, `block-shutdown-duration`, `terminationGracePeriodSeconds`, the drain settings, and the app health check (on an app-only endpoint) so the block ends early if the app genuinely dies — including how to size the durations relative to each other, a warning against probing the app through the sidecar (circular dependency), and a note that this guidance does not apply to native sidecars (where the app terminates first by design).

- `operations/hosting/kubernetes/kubernetes-production.md` — new section "Graceful shutdown for actor hosts" (main content)
- `developing-applications/building-blocks/actors/actors-runtime-config.md` — short note that the drain settings require the app to outlive the sidecar's shutdown, linking to the new section
- `operations/resiliency/health-checks/sidecar-health.md` — one-line caveat under "Delay graceful shutdown" (the flag list there is also duplicated and can be fixed in passing)

None — documentation of existing runtime behavior.

**Additional context**

Happy to submit the PR; the content is written and follows the docs contribution guide.

shaul-apiiro:
- [x] Commands include options for Linux, MacOS, and Windows within tabpane (n/a — single Kubernetes manifest, no OS-specific commands)
- [x] New file and folder names are globally unique (n/a — no new files)
- [x] Images use HTML style and have alternative text (n/a — no images)
- [x] Places where multiple code/command options are given have tabpane (n/a)

## Description

Kubernetes sends `SIGTERM` to the application container and the daprd sidecar in parallel. For actor hosts this breaks the sidecar's shutdown sequence: an idle app exits in milliseconds, so the sidecar's actor deactivations hit a dead app (connection refused, `OnDeactivateAsync` never runs) and, until the sidecar exits and placement re-disseminates, other sidecars keep routing to actors on the dying pod — failed invocations on every ordinary rolling restart. `drainOngoingCallTimeout`/`drainRebalancedActors` have nothing to act on because the app is already gone. Verified experimentally on dapr 1.18.4.

This PR documents the sidecar-outlives-app pattern:

- **`kubernetes-production.md`**: new "Graceful shutdown for actor hosts" section — the failure sequence, a working deployment example (`preStop` sleep, `block-shutdown-duration`, `terminationGracePeriodSeconds`, app health check on an app-only path), duration-sizing rules, a warning that liveness/app-health paths must not depend on the sidecar, and a notice that this guidance applies to default injection (native sidecars terminate the app first by design).
- **`actors-runtime-config.md`**: "Draining at pod termination" note — the drain settings only take effect if the app outlives the sidecar's shutdown; links to the new section.
- **`sidecar-health.md`**: caveat under "Delay graceful shutdown" for actor hosts; also removes an accidentally duplicated pair of bullets in the flag list there.

## Issue reference

Closes #5316


daprdocs/content/en/operations/hosting/kubernetes/kubernetes-production.md
```diff
@@ -382,6 +382,63 @@ If you must run a non-default UID & GID in Kubernetes, set the following env var
...
+  name: myactorapp
+spec:
+  template:
+    metadata:
+      annotations:
+        dapr.io/enabled: "true"
+        dapr.io/app-id: "myactorapp"
+        dapr.io/app-port: "8080"
+        # Hold the sidecar's shutdown open so it can drain and deactivate actors.
+        dapr.io/block-shutdown-duration: "20s"
```
JoshVanL, line 404: `block-shutdown-duration` is not needed for the problem this section fixes, and it widens the window the section opens by describing.

The `preStop` sleep on its own already solves the stated problem: the app is still serving while daprd shuts down, so both the drain and the subsequent `HaltAll` run against a live app. What the extra 20s block adds is a cost. While daprd is blocked it has not yet cancelled its runner context (`pkg/runtime/runtime.go:523-547` at v1.18.2), so the placement stream stays open, the terminating pod stays in the placement table, other sidecars keep routing to it, and new actors can still be activated on it.

That is exactly the failure described in the opening paragraph, and this adds 20 seconds of it to every pod deletion, including ordinary rolling restarts.

Suggestion: either drop `dapr.io/block-shutdown-duration` from the example (and from the sizing formula), or say explicitly what it buys on top of the `preStop` hook so readers can judge the tradeoff.

daprdocs/content/en/operations/hosting/kubernetes/kubernetes-production.md
```diff
@@ -382,6 +382,63 @@ If you must run a non-default UID & GID in Kubernetes, set the following env var
...
+                # Keep the app serving while the sidecar drains and deactivates actors.
+                command: ["/bin/sh", "-c", "sleep 30"]
+```
+
+In the application, set `drainRebalancedActors: true` and a `drainOngoingCallTimeout` at or above the p99 duration of your actor method handlers ([Actor runtime configuration parameters]({{% ref actors-runtime-config.md %}})).
+
+### Sizing the durations
+
+```
+preStop sleep                 >= block-shutdown-duration + drainOngoingCallTimeout + margin
```
JoshVanL, line 428: `margin` has a hard floor of 5 seconds, and the example's numbers fall below it.

Once the drain finishes, daprd deactivates every hosted actor via `p.table.HaltAll(cctx)` under a **hardcoded 5 second** context (`pkg/actors/internal/placement/placement.go:163` at v1.18.2, unchanged on master). That 5s is the entire budget in which every `OnDeactivateAsync` runs, and it comes *after* `drainOngoingCallTimeout`, not inside it.

So the formula needs `margin >= 5s` to be correct. With the example's numbers (block 20s, preStop 30s) `drainOngoingCallTimeout` is silently capped at 5s. A reader who follows the advice a few lines up and sets it to their handler p99 of, say, 8s gets `SIGTERM` delivered to the app part way through deactivation, and lands back in the failure this page exists to prevent, with no obvious signal.

Suggestion: make the deactivation budget explicit in the formula:

```
preStop sleep                 >= block-shutdown-duration + drainOngoingCallTimeout + 5s (actor deactivation) + margin
terminationGracePeriodSeconds >  preStop sleep + app graceful shutdown
```

daprdocs/content/en/operations/hosting/kubernetes/kubernetes-production.md
```diff
@@ -382,6 +382,63 @@ If you must run a non-default UID & GID in Kubernetes, set the following env var
...
+  template:
+    metadata:
+      annotations:
+        dapr.io/enabled: "true"
+        dapr.io/app-id: "myactorapp"
+        dapr.io/app-port: "8080"
+        # Hold the sidecar's shutdown open so it can drain and deactivate actors.
+        dapr.io/block-shutdown-duration: "20s"
+        # Let the sidecar end the block early if the app really is gone.
+        dapr.io/enable-app-health-check: "true"
```
JoshVanL, line 406: Turning on the app health check changes steady-state behaviour on actor hosts, not just shutdown behaviour, and the tightened threshold points the wrong way.

When the app reports unhealthy, daprd calls `a.actors.UnRegisterHosted(ctx, a.appConfig.Entities...)` (`pkg/runtime/runtime.go:1007` at v1.18.2). That deregisters every hosted actor type and triggers a cluster-wide placement dissemination and rebalance. Correct on a genuine shutdown, expensive on a false positive.

`dapr.io/app-health-probe-timeout` defaults to 500ms (`pkg/injector/patcher/sidecar.go:112`), and this example lowers `dapr.io/app-health-threshold` to `2` with a 3s interval. So two consecutive probes slower than 500ms, roughly a 6 second blip from a GC pause, a thread-pool stall or a momentarily slow dependency, rebalances every actor off a pod that is fine.

Suggestions:
- Leave `dapr.io/app-health-threshold` at its default instead of lowering it; a faster unhealthy signal here mostly buys faster false positives.
- Add a note that enabling app health checks on an actor host makes actor placement sensitive to probe latency, so the probe endpoint must be cheap and `dapr.io/app-health-probe-timeout` sized for it.

daprdocs/content/en/operations/hosting/kubernetes/kubernetes-production.md
```diff
@@ -382,6 +382,63 @@ If you must run a non-default UID & GID in Kubernetes, set the following env var
...
+      # Must exceed the app's preStop sleep plus the app's own graceful shutdown.
+      terminationGracePeriodSeconds: 60
+      containers:
+        - name: myactorapp
+          image: myregistry/myactorapp:latest
+          lifecycle:
+            preStop:
+              exec:
+                # Keep the app serving while the sidecar drains and deactivates actors.
+                command: ["/bin/sh", "-c", "sleep 30"]
```
JoshVanL, line 420: `/bin/sh` and `sleep` do not exist in distroless or `scratch` images.

Go and Rust actor hosts are very commonly built `FROM scratch` or on `gcr.io/distroless/static`, which is also what several Dapr actor samples produce. On those images this hook fails, Kubernetes proceeds straight to `SIGTERM`, and the whole pattern silently reverts to the broken behaviour the section describes. The only signal is a `FailedPreStopHook` event, which is easy to miss. Since this page exists precisely because the failure mode is silent, it seems worth not introducing a second silent failure in the fix.

Suggestion: lead with the native sleep handler, which needs nothing in the image (beta and on by default since Kubernetes 1.30, GA in 1.33):

```yaml
          lifecycle:
            preStop:
              sleep:
                seconds: 30
```

and keep the `exec` form as the fallback for clusters older than 1.30, noting that it requires a shell and `sleep` in the application image.

daprdocs/content/en/operations/hosting/kubernetes/kubernetes-production.md
```diff
@@ -382,6 +382,63 @@ If you must run a non-default UID & GID in Kubernetes, set the following env var
...
+```
+
+In the application, set `drainRebalancedActors: true` and a `drainOngoingCallTimeout` at or above the p99 duration of your actor method handlers ([Actor runtime configuration parameters]({{% ref actors-runtime-config.md %}})).
```
JoshVanL, line 423: This sizing rule contradicts the one in "Sizing the durations" below.

Here: set `drainOngoingCallTimeout` "at or above the p99 duration of your actor method handlers". A few lines down: keep it "below the Placement service's `disseminateTimeout`", and the section that links to recommends staying comfortably below it, "for example, 5 seconds or lower", against an 8s default.

For any workload whose handler p99 is above roughly 5 seconds, both rules cannot be satisfied. The page never says which one wins, nor that `dapr_placement.disseminateTimeout` has to be raised to accommodate a longer drain.

Suggestion: state the precedence, along the lines of "size `drainOngoingCallTimeout` to your handler p99; if that is above ~5s, raise `dapr_placement.disseminateTimeout` to match, otherwise the drain is clamped".

daprdocs/content/en/operations/hosting/kubernetes/kubernetes-production.md
```diff
@@ -382,6 +382,63 @@ If you must run a non-default UID & GID in Kubernetes, set the following env var
...
+
+{{% alert title="Important" color="warning" %}}
+Point the app health check, and any Kubernetes liveness probe on the application container, at an endpoint served by the application alone. A probe path that depends on the sidecar is circular: during shutdown the sidecar stops serving, the probe fails, and the container you are trying to keep alive gets killed.
+{{% /alert %}}
+
+{{% alert title="Native sidecars" color="primary" %}}
+This guidance applies to the default sidecar injection. With [native sidecars]({{% ref "sidecar.md#native-sidecars-kubernetes-128" %}}), Kubernetes terminates the application container first by design: the `preStop` overlap above does not apply, and actor deactivation still happens after the application has exited.
```
JoshVanL, line 439: For native sidecars this is not a caveat, it is an unavoidable failure, and the callout currently reads as the opposite.

With native sidecars daprd is an init container with `restartPolicy: Always`, so Kubernetes terminates the application container and waits for it to exit before signalling the sidecar. Actor deactivation therefore *always* runs against a dead app, and there is no `preStop` arrangement that fixes it: no ordering exists in which the app is still alive when daprd deactivates its actors.

In other words, the multi-second window of failed actor invocations described at the top of this section is unavoidable in that mode. But this is an informational (`color="primary"`) note whose first clause, "the `preStop` overlap above does not apply", reads to a `nativeSidecar: true` operator as "not affected by any of this", and "by design" suggests the outcome is intended rather than a known gap.

Suggestion: switch to `color="warning"` and lead with the consequence, that clean actor deactivation is not currently achievable when running with native sidecars, so actor hosts should stay on the default injection until that is addressed.

daprdocs/content/en/developing-applications/building-blocks/actors/actors-runtime-config.md
```diff
@@ -42,6 +42,10 @@ The clamping only protects daprd from resetting its own placement stream. A drai
 
 Most Dapr SDKs leave `drainOngoingCallTimeout` unset unless your application configures it, so the 2-second default applies. The .NET Actors.Next SDK is an exception: it sets a 30-second default, which sits exactly at the clamp boundary; configure a lower value explicitly.
 
+## Draining at pod termination
+
+`drainOngoingCallTimeout` and `drainRebalancedActors` only take effect while the sidecar can still reach your application. On Kubernetes both containers receive `SIGTERM` at the same time, so an application that exits promptly is gone before the sidecar drains, leaving these settings nothing to act on. Configure the application container to outlive the sidecar's shutdown: see [Graceful shutdown for actor hosts]({{% ref "kubernetes-production.md#graceful-shutdown-for-actor-hosts" %}}).
```
JoshVanL, line 47: "On Kubernetes both containers receive `SIGTERM` at the same time" is stated unconditionally, and it is false under native sidecars, where Kubernetes terminates the application container first and waits for it.

The Kubernetes page carries that qualification in its own callout, but this is the actor runtime config reference and is linked from a lot of places. Readers will take the timing model from this sentence without following the link, and then size `drainOngoingCallTimeout` on a premise that does not hold in their cluster.

Suggestion: "On Kubernetes, with the default sidecar injection, both containers receive `SIGTERM` at the same time ...", with the native sidecar difference either mentioned here or clearly deferred to the linked section.

daprdocs/content/en/operations/hosting/kubernetes/kubernetes-production.md
```diff
@@ -382,6 +382,63 @@ If you must run a non-default UID & GID in Kubernetes, set the following env var
...
 ```
 
+## Graceful shutdown for actor hosts
+
+When a pod is deleted, Kubernetes sends `SIGTERM` to the application container and the `daprd` sidecar at the same time. An idle application often exits within milliseconds — before the sidecar shuts down its actor runtime. The sidecar then fails to deactivate its hosted actors (connection refused; `OnDeactivateAsync` never runs), and until it exits and the placement table is re-disseminated, other sidecars keep routing calls to actors on the dying pod. Every ordinary rolling restart produces a multi-second window of failed actor invocations.
+
+The remedy is to keep the application serving while the sidecar drains and deactivates its actors. Use a `preStop` hook on the *application* container, hold the sidecar's shutdown open with `block-shutdown-duration`, and size the termination grace period to cover both. A working starting point — tune the durations to your workload:
+
+```yaml
+apiVersion: apps/v1
```
JoshVanL, line 392: This manifest cannot be applied as written: `apps/v1` Deployments require `spec.selector`, and it must match `spec.template.metadata.labels`. Both are missing, so `kubectl apply` fails with `spec.selector: Required value`.

It is introduced as "a working starting point" and is otherwise complete enough (apiVersion, kind, metadata, container image) that readers will copy it straight into a file and apply it.

Suggestion:

```yaml
spec:
  selector:
    matchLabels:
      app: myactorapp
  template:
    metadata:
      labels:
        app: myactorapp
      annotations:
        dapr.io/enabled: "true"
        ...
```

JoshVanL (changes requested):

daprdocs/content/en/operations/hosting/kubernetes/kubernetes-production.md
shaul-apiiro: Dropped from the example and the formula in 57de5c1 — preStop alone covers the stated problem.

daprdocs/content/en/operations/hosting/kubernetes/kubernetes-production.md
shaul-apiiro: Adopted your formula (explicit 5s deactivation term) and noted deactivation runs after the drain, in 57de5c1.

daprdocs/content/en/operations/hosting/kubernetes/kubernetes-production.md
shaul-apiiro: Reworked in 57de5c1: annotations removed from the example; app health check now covered by a note — steady-state rebalance effect, cheap probe endpoint, sized probe-timeout, default threshold.

daprdocs/content/en/operations/hosting/kubernetes/kubernetes-production.md
shaul-apiiro: Switched to the native sleep handler, with the exec form noted as a pre-1.30 fallback and its distroless caveat, in 57de5c1.

daprdocs/content/en/operations/hosting/kubernetes/kubernetes-production.md
shaul-apiiro: Precedence stated in 57de5c1: size to handler p99, raise dapr_placement.disseminateTimeout when above a few seconds — with the stream-reset (not clamp) consequence.

daprdocs/content/en/operations/hosting/kubernetes/kubernetes-production.md
shaul-apiiro: Changed to a warning leading with the consequence (clean deactivation not achievable; keep actor hosts on default injection) in 57de5c1.

daprdocs/content/en/developing-applications/building-blocks/actors/actors-runtime-config.md
shaul-apiiro: Qualified to default injection, with the native-sidecar ordering stated, in 57de5c1.

daprdocs/content/en/operations/hosting/kubernetes/kubernetes-production.md
shaul-apiiro: Added spec.selector and matching labels in 57de5c1.

daprdocs/content/en/operations/resiliency/health-checks/sidecar-health.md
```diff
@@ -135,10 +135,10 @@ Dapr accepts a [`dapr.io/block-shutdown-duration` annotation or `--dapr-block-sh
 
 During this period, all subscriptions and input bindings are closed. This is useful for applications that need to use the Dapr APIs as part of their own shutdown procedure.
 
+For actor hosts on Kubernetes, the block only helps if the application outlives it *and* the actor drain and deactivation that follow — keep the app container alive with a `preStop` hook sized for all three. See [Graceful shutdown for actor hosts]({{% ref "kubernetes-production.md#graceful-shutdown-for-actor-hosts" %}}).
```
JoshVanL, line 138: This now points at a formula that has two terms, not three, and the case the block exists for has lost its guidance.

The linked section's "Sizing the durations" reads `preStop sleep >= drainOngoingCallTimeout + 5s (actor deactivation) + margin`, and the example no longer sets `dapr.io/block-shutdown-duration` at all. A reader sent here for a hook "sized for all three" finds two, and no mention of the annotation they just read about.

There is a substantive gap behind the wording too. Without the block, daprd finishes drain plus deactivation and exits roughly 7 to 12 seconds into a 30 second `preStop`, so the app's own graceful shutdown, which the `terminationGracePeriodSeconds` formula explicitly budgets for, always runs with a dead sidecar. That is precisely the case this page introduces the block for: "applications that need to use the Dapr APIs as part of their own shutdown procedure". After this PR nothing tells those readers how to combine the two.

Suggestion: keep the block as an optional term on the Kubernetes page, for example

```
preStop sleep >= block-shutdown-duration + drainOngoingCallTimeout + 5s (actor deactivation) + margin
```

with a line saying the block term is only needed when the application calls Dapr APIs during its own shutdown. Then "all three" here is accurate. Otherwise drop "and the actor drain and deactivation that follow ... sized for all three" in favour of wording that matches the two-term formula.

daprdocs/content/en/operations/hosting/kubernetes/kubernetes-production.md
```diff
@@ -382,6 +382,72 @@ If you must run a non-default UID & GID in Kubernetes, set the following env var
...
+On Kubernetes older than 1.30, use the `exec` form instead of the native `sleep` handler — it requires a shell and `sleep` in the application image, so it silently does nothing on distroless or `scratch` images:
+
+```yaml
+          lifecycle:
+            preStop:
+              exec:
+                command: ["/bin/sh", "-c", "sleep 30"]
+```
+
+In the application, set `drainRebalancedActors: true` and size `drainOngoingCallTimeout` to the p99 duration of your actor method handlers; if that exceeds a few seconds, raise the Placement service's `disseminateTimeout` above it with headroom for acknowledging the placement update, otherwise Placement resets streams that outlive it ([Drain timeout clamping]({{% ref "actors-runtime-config.md#drain-timeout-clamping" %}})).
```
JoshVanL, line 430: Raising the Placement service's `disseminateTimeout` is only half of it: daprd clamps the drain against its own dissemination timeout as well.

`ClampDrainOngoingCallTimeout` (`pkg/actors/api/config.go:98`, called from `pkg/actors/actors.go:433`) reduces the effective drain to 80% of daprd's `actors-disseminate-timeout` whenever the configured value meets or exceeds it. That timeout defaults to 30s (`DefaultActorsDisseminationTimeout`, `pkg/runtime/config.go:76`) and is set with `dapr.io/actors-disseminate-timeout`.

So a reader with a 40s handler p99 who follows this sentence, raises `dapr_placement.disseminateTimeout` to 60s and sets `drainOngoingCallTimeout: 40s`, still gets an effective drain of 24s, with only a log warning to say so. The section you link to covers both timeouts, but this sentence names one as though it were the only constraint, and it is the sentence most readers will act on.

Suggestion: "... raise both the Placement service's `disseminateTimeout` and daprd's own `dapr.io/actors-disseminate-timeout` above it ...", or add "above 30s you must also raise `dapr.io/actors-disseminate-timeout`, or daprd clamps the drain to 80% of it".

daprdocs/content/en/operations/hosting/kubernetes/kubernetes-production.md
```diff
@@ -382,6 +382,72 @@ If you must run a non-default UID & GID in Kubernetes, set the following env var
...
+
+### Sizing the durations
+
+```
+preStop sleep                 >= drainOngoingCallTimeout + 5s (actor deactivation) + margin
+terminationGracePeriodSeconds >  preStop sleep + app graceful shutdown
+```
+
+Actor deactivation runs *after* the drain, under a fixed 5-second budget, so the `preStop` sleep must cover both. If `terminationGracePeriodSeconds` expires first, Kubernetes sends `SIGKILL` and you are back to the original problem.
```
JoshVanL, line 439: Worth saying that the 5 seconds covers every hosted actor collectively, and that nothing extends it.

`HaltAll` fans out across all actor factories concurrently under a single `context.WithTimeout(context.Background(), time.Second*5)` (`pkg/actors/internal/placement/placement.go:161`, `pkg/actors/table/table.go` `HaltAll`). It is one budget for the whole host, not per actor, and there is no annotation or config that changes it.

As written, "a fixed 5-second budget, so the `preStop` sleep must cover both" reads as though sizing the hook correctly is sufficient for deactivation. On a host with a large actor population, or with an `OnDeactivateAsync` that persists state, the budget is exceeded collectively and the remaining actors are never deactivated, no matter how generous the `preStop` sleep and grace period are. Readers who hit this will otherwise assume they mis-sized something.

Suggestion: "Actor deactivation runs *after* the drain, under a single fixed 5-second budget shared by every actor the host owns, so the `preStop` sleep must cover both. The budget is not configurable: hosts with very large actor populations, or slow deactivation handlers, may not finish within it."

daprdocs/content/en/operations/hosting/kubernetes/kubernetes-production.md
```diff
@@ -382,6 +382,72 @@ If you must run a non-default UID & GID in Kubernetes, set the following env var
...
+      # Must exceed the app's preStop sleep plus the app's own graceful shutdown.
+      terminationGracePeriodSeconds: 60
+      containers:
+        - name: myactorapp
+          image: myregistry/myactorapp:latest
+          lifecycle:
+            preStop:
+              # Keep the app serving while the sidecar drains and deactivates actors.
+              sleep:
+                seconds: 30
```
JoshVanL, line 418: Where does 30 come from? The section's own formula gives about 10 for the defaults this example implies.

The manifest sets no `drainOngoingCallTimeout`, so the 2s default applies, and `preStop sleep >= drainOngoingCallTimeout + 5s + margin` works out at roughly 10s. Copied verbatim, this makes every pod deletion take at least 30 seconds and stretches rolling restarts to match, with about 20 of those seconds spent with the app idle and its sidecar already exited.

That is harmless but it is three times the cost the formula asks for, and a reader cannot tell whether 30 encodes an assumption they should copy or is just a round number.

Suggestion: either state the assumption in the comment, for example "assumes `drainOngoingCallTimeout: 20s`; see Sizing the durations below", or use a value the formula produces for the defaults, such as 15.

JoshVanL (approved):

daprdocs/content/en/operations/resiliency/health-checks/sidecar-health.md
shaul-apiiro: Added the optional block term to the formula with the when-it-applies note, in f94ce18 — "all three" here is accurate again.

daprdocs/content/en/operations/hosting/kubernetes/kubernetes-production.md
shaul-apiiro: Added your suggested clause on dapr.io/actors-disseminate-timeout and the 80% clamp, in f94ce18.

daprdocs/content/en/operations/hosting/kubernetes/kubernetes-production.md
shaul-apiiro: Adopted: single fixed 5s budget shared by every actor, not configurable, may not finish on large populations — in f94ce18.

daprdocs/content/en/operations/hosting/kubernetes/kubernetes-production.md
shaul-apiiro: Changed to 15 (matches the formula for the example's defaults), in f94ce18.

PR #5325, opened 2026-09-18
[Backport v1.18] Correct Redis pub/sub redeliverInterval and processingTimeout reference

### LINKED ISSUE #5324: Redis pub/sub reference: redeliverInterval and processingTimeout defaults are swapped, and redeliverInterval does not accept a duration string
## Describe the issue

The Redis pub/sub reference page documents two things about `redeliverInterval` and `processingTimeout` that do not match the component's behaviour. The affected rows are byte-identical on v1.17, v1.18 and v1.19.

Code links below are pinned to the `v1.18.4` component release.

**1. The two defaults are swapped.**

The page states `redeliverInterval` defaults to `"60s"` and `processingTimeout` to `"15s"`. The component sets the opposite, in [`common/component/redis/redis.go#L137-L138`](https://github.com/dapr/components-contrib/blob/58a792451a050605301ff05cdff5e6ff2ad79143/common/component/redis/redis.go#L137-L138):

```go
settings.ProcessingTimeout = 60 * time.Second
settings.RedeliverInterval = 15 * time.Second
```

**2. `redeliverInterval` does not accept a Go duration string.**

The page says it "Can use either be Go duration string (for example "ms", "s", "m") or milliseconds number", and gives `"30s"` as its first example. Only a bare milliseconds number is honoured.

`Settings.RedeliverInterval` is tagged `mapstructure:"-"` ([`settings.go#L109`](https://github.com/dapr/components-contrib/blob/58a792451a050605301ff05cdff5e6ff2ad79143/common/component/redis/settings.go#L109)), so the decoder never populates it. The only path that sets it is a `strconv.ParseUint` of the raw value as milliseconds ([`redis.go#L159-L166`](https://github.com/dapr/components-contrib/blob/58a792451a050605301ff05cdff5e6ff2ad79143/common/component/redis/redis.go#L159-L166)). A duration string fails that parse, the error is discarded on the assumption the decoder already handled it, and the value silently stays at the default — with no warning.

Calling `ParseClientFromProperties` directly:

```
redeliverInterval="30s"    -> 15s   (ignored)
redeliverInterval="30000"  -> 30s
```

`processingTimeout` is documented identically but is tagged `mapstructure:"processingTimeout"` ([`settings.go#L111`](https://github.com/dapr/components-contrib/blob/58a792451a050605301ff05cdff5e6ff2ad79143/common/component/redis/settings.go#L111)), so it does accept both forms. The two adjacent settings behave differently.

This one is worth prioritising: the page instructs people to use a form that silently does nothing, and gives it as the lead example, so anyone following the docs ends up with a redelivery interval they did not choose.

The component-side fix is tracked in dapr/components-contrib#4600. Until it ships, the page should describe what the released code accepts.

## Affected pages

`daprdocs/content/en/reference/components-reference/supported-pubsub/setup-redis-pubsub.md`

## Release Note

RELEASE NOTE: **UPDATE** Correct the documented defaults and accepted formats for the Redis pub/sub `redeliverInterval` and `processingTimeout` settings.

MyMirelHub:
Backport of #5326 to `v1.18`.

The affected rows are byte-identical across v1.17, v1.18 and v1.19, so the change is the same on each branch.

Fixes #5324. (The closing keyword sits here rather than on #5326 because GitHub only auto-closes from the default branch.)


nelson-parente (approved):

marcduiker (approved):
LGTM

PR #5326, opened 2026-09-18
Correct Redis pub/sub redeliverInterval and processingTimeout reference

MyMirelHub:
Tracked in #5324. Backported to v1.18 in #5325 and v1.17 in #5327.

Two corrections to the Redis pub/sub reference.

**The defaults were swapped.** The page stated `redeliverInterval` defaults to `"60s"` and `processingTimeout` to `"15s"`. The component sets the opposite — `ProcessingTimeout = 60 * time.Second`, `RedeliverInterval = 15 * time.Second` — unchanged across v1.17.7 and v1.18.4.

**`redeliverInterval` does not accept a Go duration string.** The page said it accepts one and gave `"30s"` as its first example. `Settings.RedeliverInterval` is tagged `mapstructure:"-"`, so the decoder never populates it and only the bare-milliseconds path applies; a duration string fails that parse, the error is discarded, and the value silently falls back to the default:

```
redeliverInterval="30s"    -> 15s   (ignored)
redeliverInterval="30000"  -> 30s
```

`processingTimeout` is documented identically but is tagged correctly and does accept both forms, so its wording is left intact apart from the default. The `redeliverInterval` row now simply states the format it accepts rather than describing the discrepancy.

The component-side fix is tracked in dapr/components-contrib#4600. Once that ships, the `redeliverInterval` wording can go back to accepting both forms; until then the page should describe what the released code does.


nelson-parente (approved):

marcduiker (approved):
LGTM

PR #5328, opened 2026-09-18
Document Redis pub/sub entryKeepAliveInterval, restore redeliverInterval duration form

MyMirelHub:
Docs for [#4601:](https://github.com/dapr/components-contrib/pull/4601)

**`entryKeepAliveInterval` (new).** The component currently lets a message's idle time run from the moment it is read from the stream, while the handler deadline only starts when a worker picks it up. Messages waiting in the sidecar's buffer therefore age towards `processingTimeout` untouched and can be redelivered while the first delivery is still running. [#4601](https://github.com/dapr/components-contrib/pull/4601) keeps the entries the sidecar is holding alive, and `entryKeepAliveInterval` controls how often; `0` restores the current behaviour.

**`redeliverInterval` wording restored.** #5326 corrected this row to say milliseconds only, which is accurate for every released version — `Settings.RedeliverInterval` is tagged `mapstructure:"-"`, so a duration string is silently discarded. [#4601 ](https://github.com/dapr/components-contrib/pull/4601) gives it a real tag, so both forms work again and the row goes back to matching `processingTimeout`. The corrected `15s` default is kept.


alicejgibbons (approved):
lgtm

